PR #21, opened 2020-10-07
Annotation management

### LINKED ISSUE #15: annotation generation and management
**Proposal** :

- https://docs.google.com/document/d/1F-j2wTN_f1KWWfp3LpwO5KTJDBIhdSwNkZQ6LUAW-pk/edit
- https://docs.google.com/spreadsheets/d/1dJwIMEJoK-CiR1JqlmOcTnnf68mc8k-AooLXL-QVtP4/edit?ouid=117619961749361969396&usp=sheets_home&ths=true

test_annotations.py:

```python
from ChildProject.projects import ChildProject, RecordingProfile
from ChildProject.annotations import AnnotationManager
import pandas as pd
import numpy as np
import os

def test_import():
    project = ChildProject("examples/valid_raw_data")
    project.import_data("output/annotations")
    project = ChildProject("output/annotations")
    am = AnnotationManager(project)

    input_annotations = pd.read_csv('examples/valid_raw_data/raw_annotations/input.csv')

    am.import_annotations(input_annotations)
    
    assert am.annotations.shape[0] == input_annotations.shape[0], "imported annotations length does not match input"

    assert all([
        os.path.exists(os.path.join(project.path, 'annotations', f))
        for f in am.annotations['annotation_filename'].tolist()
    ]), "some annotations are missing"

    segments = am.annotations['annotation_filename'].map(lambda f: pd.read_csv(os.path.join(project.path, 'annotations', f))).tolist()
    segments = pd.concat(segments)
    CHI_segments = segments[(segments['speaker_type'] == 'CHI') & (segments['ling_type'] == 1)]
    
    assert np.isclose((CHI_segments['segment_offset']-CHI_segments['segment_onset']).sum(), 2.867) == True
```

Generation ?

```python
from ChildProject.projects import ChildProject, RecordingProfile
from ChildProject.VTCAnnotation import VTCAnnotation
import pandas as pd
import os

def test_vtc():
    project = ChildProject("examples/valid_raw_data")
    project.import_data("output/vtc")
    project.convert_recordings(RecordingProfile(name = '16kHz', sampling = 16000))

    inputs = [
        pd.read_csv('examples/vtc_input1.csv'),
        pd.read_csv('examples/vtc_input2.csv')
    ]

    vtc = VTCAnnotation(project)

    for input in inputs:
        vtc.pre_process('vtc_test', input, '16kHz')
        vtc.process(options = ['--device=gpu'])
        vtc.post_process()

    index = pd.read_csv("output/vtc/annotations/vtc_test/index.csv")
    
    assert sum(input.shape[0] for input in inputs) == index.shape[0], "input index is not complete"
    assert all(index['completed'].tolist()), "some task(s) failed to complete"
    assert all([
        os.path.exists(os.path.join("output/vtc/annotations/vtc_test/output", f))
        for f in index['output_filename'].tolist()
    ]), "output files are missing"
```

lucasgautheron:
Closes #15

- [x] proper tests against desired output
- [x] textgrid support
- [x] rttm support `awaiting tests and validation`
- [x] eaf support
- [x] error handling (at the moment, if some file leads to a crash, the whole process crashes)
- [x] scripts
- [x] documentation


ChildProject/annotations.py
```diff
@@ -20,6 +20,20 @@ class AnnotationManager:
         IndexColumn(name = 'format', description = 'input annotation format', regex = r"(TextGrid|eaf|rttm)", required = True)
     ]
 
+    SEGMENTS_COLUMNS = [
+        IndexColumn(name = 'annotation_file', description = 'raw annotation path relative to /raw_annotations/', required = True),
+        IndexColumn(name = 'segment_onset', description = 'segment start time in seconds', regex = r"[0-9]{1,}(\.[0-9]{3})?", required = True),
+        IndexColumn(name = 'segment_offset', description = 'segment end time in seconds', regex = r"[0-9]{1,}(\.[0-9]{3})?", required = True),
+        IndexColumn(name = 'speaker_id', description = '', required = True),
```
lucasgautheron, line 27: Can you please help me fill empty descriptions here ? :) @alecristia

examples/valid_raw_data/raw_annotations/example.TextGrid
```diff
@@ -0,0 +1,158 @@
...
+            xmin = 5.209750000000014 
+            xmax = 10 
+            text = "" 
+    item [5]:
+        class = "IntervalTier" 
+        name = "    FA1" 
+        xmin = 0 
+        xmax = 10 
+        intervals: size = 1 
+        intervals [1]:
```
lucasgautheron, line 87: should we discard all intervals with text == "" ? Is it safe to do so ? @alecristia

tests/test_annotations.py
```diff
@@ -0,0 +1,37 @@
...
+    assert am.annotations.shape[0] == input_annotations.shape[0], "imported annotations length does not match input"
+
+    assert all([
+        os.path.exists(os.path.join(project.path, 'annotations', f))
+        for f in am.annotations['annotation_filename'].tolist()
+    ]), "some annotations are missing"
+
+    errors, warnings = am.validate()
+    assert len(errors) > 0 or len(warnings) > 0, "malformed annotations detected"
+
```
lucasgautheron, line 28: here, it would be good to test converted annotations agains the desired output in some way ! @alecristia

ChildProject/annotations.py
```diff
@@ -0,0 +1,218 @@
...
+        return pd.DataFrame(segments)
+
+    def load_eaf(self, filename):
+        path = os.path.join(self.project.path, 'raw_annotations', filename)
+        eaf = pympi.Elan.Eaf(path)
+
+
+        return None
+
+    def load_vtc_rttm(self, filename):
```
lucasgautheron, line 160: is the logic correct here ?
shouldn't CHI be translated into OCHI and KCHI in CHI or something of this sort ?

examples/valid_raw_data/raw_annotations/example.TextGrid
alecristia: yes, that's right! those are created by praat to fill in the blanks

ChildProject/annotations.py
alecristia: done!

examples/valid_raw_data/raw_annotations/example.TextGrid
alecristia: yes, these are gaps filled in by praat

tests/test_annotations.py
alecristia: are you asking me to list the tests? I'd be happy to - I know many ways in which these can be broken!

lucasgautheron: Sure :)

examples/valid_raw_data/raw_annotations/example.TextGrid
alecristia: yes!

ChildProject/annotations.py
alecristia: done)

ChildProject/annotations.py
```diff
@@ -0,0 +1,218 @@
...
+
+            segments.read()
+            res = segments.validate()
+            errors += res[0]
+            warnings += res[1]
+
+        return errors, warnings
+        
+
+    def load_textgrid(self, filename):
```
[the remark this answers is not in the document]
alecristia: Only one question:
SPEAKER_ID_TO_TYPE -- so we'll generate the types (line 141 in the next chunk)? that's what we did in the past with exactly the code you see there. BUT notice that this code ignores far away speech (LF2P) and overlapping speech (2POPMT), which we don't want to lose. (In the past Marvin dropped it because he was training a model, so he needed clean speech.) This also removes Noise (a tier called noise) but this was annotated in such an erratic way that we can probably forget it. (I'd also rewrite Marvin's code with regular expressions, but I know that's individual choice).
Otherwise this bit looks great.

ChildProject/annotations.py
alecristia: the logic is perfect.
You're right, Marvin's VTC uses KCHI for key child and CHI for other children (confusingly, since everyone else in the language development world uses CHI for the key child). So yes, you'd need to add to the section starting line 37, with the dictionary, a translation CHI -> OCH (notice no I); and KCHI -> CHI

examples/valid_raw_data/raw_annotations/example.TextGrid
alecristia: oops I guess my comments WERE going through. Funny how my replies get shorter and shorter...

ChildProject/annotations.py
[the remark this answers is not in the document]
lucasgautheron: What do you think is the desired behavior in this case ? Right now, the information is not lost at least, but SPEAKER TYPE is NA when ID = LF2P or 2POPMT

ChildProject/annotations.py
```diff
@@ -0,0 +1,287 @@
...
+                    'mwu_type': 'NA',
+                    'addresseee': 'NA',
+                    'transcription': 'NA'
+                }
+
+                segments.append(segment)
+
+        return pd.DataFrame(segments)
+
+    def load_eaf(self, filename):
```
lucasgautheron, line 153: It passes the tests, but can you see any issue from looking at the code ?
@alecristia

ChildProject/annotations.py
```diff
@@ -0,0 +1,287 @@
...
+                    'segment_onset': "{:.3f}".format(start_t/1000),
+                    'segment_offset': "{:.3f}".format(end_t/1000),
+                    'speaker_id': tier_name,
+                    'ling_type': 'NA',
+                    'speaker_type': self.SPEAKER_ID_TO_TYPE[tier_name] if tier_name in self.SPEAKER_ID_TO_TYPE else 'NA',
+                    'vcm_type': 'NA',
+                    'lex_type': 'NA',
+                    'mwu_type': 'NA',
+                    'addresseee': 'NA',
+                    'transcription': value if value != '0.' else '0'
```
lucasgautheron, line 176: Should we really translate 0. to 0 ? This was necessary for the tests to pass

ChildProject/annotations.py
```diff
@@ -122,6 +130,17 @@ def load_textgrid(self, filename):
         path = os.path.join(self.project.path, 'raw_annotations', filename)
         textgrid = pympi.Praat.TextGrid(path)
 
+        def ling_type(s):
```
lucasgautheron, line 133: is this good for you ? @alecristia

ChildProject/annotations.py
```diff
@@ -77,6 +78,13 @@ class AnnotationManager:
         'UC6': 'OCH'
     }
 
+    VTC_SPEAKER_TYPE_TRANSLATION = defaultdict(lambda: 'NA', {
```
lucasgautheron, line 81: does this mapping seem alright to you ? @alecristia

ChildProject/annotations.py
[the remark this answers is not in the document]
alecristia: this is indeed the right behavior! If Marvin or someone else wants to retrain, they can select down by speaker type removing NAs

ChildProject/annotations.py
alecristia: this is a mistake in the google spreadsheet, which autocorrected 0 to 0., when in fact,  the correct entry for transcription is 0. (with the stop). So if anything, we should do the opposite: it's a common error for humans to forget to write the stop.

ChildProject/annotations.py
alecristia: 'segment_onset': "{:.3f}".format(interval[0]) <-- note that in theory VTC should return 2 digits (sometimes it does not, but I don't know why, since frames are 10 ms long)

Also, this is a good place to fix entries by humans that go beyond the period they needed to code -- so lines 137-8 would fix onset/offsets that go beyond the coding period.

Otherwise, it looks good!

lucasgautheron: Thanks!

tests/test_annotations.py
lucasgautheron: I do not think it is useful to differentiate them for now :)

ChildProject/annotations.py
```diff
@@ -0,0 +1,306 @@
...
+        'FA0': 'FEM',
+        'FA1': 'FEM',
+        'FA2': 'FEM',
+        'FA3': 'FEM',
+        'FA4': 'FEM',
+        'FA5': 'FEM',
+        'FA6': 'FEM',
+        'FA7': 'FEM',
+        'FA8': 'FEM',
+        'FAE': 'ELE',
```
alecristia, line 54: line 54 should be removed

ChildProject/annotations.py
```diff
@@ -0,0 +1,306 @@
...
+        IndexColumn(name = 'addresseee', description = 'T if target-child-directed, C if other-child-directed, A if adult-directed, U if uncertain or other', required = True),
+        IndexColumn(name = 'transcription', description = 'orthographic transcription of the speach', required = True)
+    ]
+
+    SPEAKER_ID_TO_TYPE = {
+        'C1': 'OCH',
+        'C2': 'OCH',
+        'CHI': 'CHI',
+        'CHI*': 'CHI',
+        'EE1': 'ELE',
```
alecristia, line 44: line 44 should be removed (ELE is electronic speech, we don't analyze it as baby input, and we don't use it for training, so this should map onto the default NA)

ChildProject/annotations.py
```diff
@@ -0,0 +1,306 @@
...
+        'FC1': 'OCH',
+        'FC2': 'OCH',
+        'FC3': 'OCH',
+        'MA0': 'MAL',
+        'MA1': 'MAL',
+        'MA2': 'MAL',
+        'MA3': 'MAL',
+        'MA4': 'MAL',
+        'MA5': 'MAL',
+        'MAE': 'ELE',
```
alecristia, line 64: line 64 should be removed

ChildProject/annotations.py
alecristia: yes, other than those 4 changes (ELE+OCHI)

ChildProject/annotations.py
```diff
@@ -0,0 +1,306 @@
...
+        'UC1': 'OCH',
+        'UC2': 'OCH',
+        'UC3': 'OCH',
+        'UC4': 'OCH',
+        'UC5': 'OCH',
+        'UC6': 'OCH'
+    }
+
+    VTC_SPEAKER_TYPE_TRANSLATION = defaultdict(lambda: 'NA', {
+        'CHI': 'OCHI',
```
alecristia: line 82 should map CHI to OCH (not OCHI)

ChildProject/annotations.py
alecristia: interesting syntax! does this mean:

- check if s is a string (and NOT convert into a string)
- if false, then return NA

What may or may not be a string? In r this depends on how you read something: in read.table, fields that have only numbers will be some form of numeric; otherwise they'll be factors -- so nothing would be string.

Examples of human coding errors:

- "1" (not a string, right? if yes, handled by l. 138-10)
- "0" (idem, both are correct)
-  " 0" (error entering it, but we get what coder meant - "0"; and this would be a string so not NAd?)
-  "¨0" (error entering it, but we get what coder meant - "0"; and this would be a string so not NAd?)
- "00" (error entering it, but we get what coder meant - "0"; and this is handled by lines 138-142 - to which I'll return)
- "x" (impossible to tell what coder meant ; and this is handled by lines 138-142)
- "01" (impossible to tell what coder meant ; and this is not yet handled by lines 138-142)

--> suggested correction: substitute out all non [0-9], so that only [0-9] remain in s

lines 138-142:

- define two boolean vars, a & b, which check whether s contains 0 or 1 -- note: in R, when a variable is FALSE, all matches are accepted (ie FALSE contains 0) - I'm guessing this doesn't happen in py?
- if either a or b are true then return a 0 -- return 0 if a is true, and otherwise 1
- BUT this means that "01" gets a "0", but we can't be sure -- so can we make a^b disjunctive? or add a line so that when both are true s in NA?

lucasgautheron: mmh, i think the code is doing what you're saying ?
Here is an example 👍 
```python
def ling_type(s):
    s = str(s)
    if not s:
        return "NA"

    a, b = ('0' in s, '1' in s)
    if a^b:
        return '0' if a else '1' 
    else:
        return 'NA'

values = ['1', '0', ' 0', '¨0', 'x', '01', '', 1]

print("\n".join([
    "ling_type('{}') = {}".format(s, ling_type(s))
    for s in values
]))
```

Output:
```
ling_type('1') = 1
ling_type('0') = 0
ling_type(' 0') = 0
ling_type('¨0') = 0
ling_type('x') = NA
ling_type('01') = NA
ling_type('') = NA
ling_type('1') = 1
```

and I've used a^b because it is disjoint indeed:

| a | b | a^b |
|---|---|-----|
| 0 | 0 | 0   |
| 0 | 1 | 1   |
| 1 | 0 | 1   |
| 1 | 1 | 0   |

so that ambiguous values such as 01 return NA :)

ChildProject/annotations.py
lucasgautheron: fixed in 0d18ce91d916924188d0b1cc2b4ea048bbc893e4

ChildProject/annotations.py
lucasgautheron: b37736009ef5ca02b1c9bc226750d3f937a8c445

ChildProject/annotations.py
lucasgautheron: fixed in b37736009ef5ca02b1c9bc226750d3f937a8c445

ChildProject/annotations.py
lucasgautheron: fixed in b37736009ef5ca02b1c9bc226750d3f937a8c445

lucasgautheron:
What should be the converted output for this :

```bash
SPEAKER namibie_aiku_20160714_1 1 1982.193 0.299 <NA> <NA> SPEECH <NA> <NA>
SPEAKER namibie_aiku_20160714_1 1 1983.496 5.496 <NA> <NA> SPEECH <NA> <NA>
SPEAKER namibie_aiku_20160714_1 1 1984.136 0.857 <NA> <NA> KCHI <NA> <NA>
SPEAKER namibie_aiku_20160714_1 1 1984.168 2.344 <NA> <NA> CHI <NA> <NA>
SPEAKER namibie_aiku_20160714_1 1 1985.4920000000002 3.4589999999999996 <NA> <NA> FEM <NA> <NA>
```

(In particular, how to handle SPEECH segments ?)

ChildProject/annotations.py
alecristia: perfect!

PR #37, opened 2020-10-15
attempt to fix eaf importation failing on python 3.5

lucasgautheron:


lucasgautheron:
Clues:

- the scripts fails on different annotations every time (not determistic)
- same version of pympi across all tested versions of python

PR #42, opened 2020-10-21
multithread annotation importation

lucasgautheron:
It is significantly faster this way... but what about thread-safety ?

### DIFF
--- a/ChildProject/annotations.py
+++ b/ChildProject/annotations.py
@@ -1,6 +1,5 @@
 from collections import defaultdict
 import datetime
-from functools import partial
 import multiprocessing as mp
 from numbers import Number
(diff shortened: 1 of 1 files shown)


ChildProject/annotations.py
```diff
@@ -268,9 +273,10 @@ def import_annotation(self, annotation, output_filename):
                 raise ValueError("file format '{}' unknown for '{}'".format(annotation_format, raw_filename))
         except Exception as e:
             self.errors.append(str(e))
```
lucasgautheron, line 275: not thread-safe. instead, we should have a error column that is filled when importation fails, and we should return the annotation row here instead of appending errors.

PR #44, opened 2020-10-27
adding data fields

lucasgautheron:


ChildProject/projects.py
```diff
@@ -156,8 +159,8 @@ class ChildProject:
         IndexColumn(name = 'its_filename', description = 'its_filename', filename = True),
         IndexColumn(name = 'upl_filename', description = 'upl_filename', filename = True),
         IndexColumn(name = 'lena_id', description = ''),
-        IndexColumn(name = 'age', description = 'age in months (rounded)', regex = '^[0-9]+$'),
-        IndexColumn(name = 'notes', description = 'free-style notes about individual recordings (avoid tabs and newlines)')
+        IndexColumn(name = 'notes', description = 'free-style notes about individual recordings (avoid tabs and newlines)'),
+        IndexColumn(name = 'daytime', description = 'yes (Y) means recording launched such that most or all of the audiorecording happens during daytime; no (N) means at least 30% of the recording may happen at night', regex = '(Y|N)')
```
lucasgautheron, line 163: We could add consistency check for that. But it requires a definition of "daytime" in terms of time range.
But in any case, it would not hurt to move it to recordings right ? @alecristia

lucasgautheron: I guess the question is : should we move it or delete it ?

alecristia (approved):

PR #95, opened 2021-01-18
child-project convert --channel option

lucasgautheron:
Paving the way for the BabyLogger, one - baby - step at a time.


lucasgautheron:
I am not sure about that, but it looks like maybe, this leaves the amount of channels the same - just sets all the channels except for the one set with --channel to zeros. But this would mean the output is as large as the original, which would be very inefficient. This needs to be checked.

Issue #109, opened 2021-02-03
Zooniverse use cases


lucasgautheron:
<table class="tg">
<thead>
  <tr>
    <th class="tg-fymr">Project</th>
    <th class="tg-7btt">Dataset</th>
    <th class="tg-fymr">Sampling algorithm</th>
    <th class="tg-fymr">Classification task(s)</th>
    <th class="tg-fymr">Launch Date</th>
    <th class="tg-fymr">URL</th>
  </tr>
</thead>
<tbody>
  <tr>
    <td class="tg-0pky">Marvin's pilot</td>
    <td class="tg-c3ow">LENA/BabyLogger data</td>
    <td class="tg-c3ow">i) for each recorded child, extract 10 sections of 30 seconds that are NOT silence (according to simple loudness detector) 
ii) cut these sections up into 500ms clips. NOTE : Loudness detector score should be averaged across LENA and BabyLogger </td>
    <td class="tg-c3ow">child voc, female adult speech, male adult speech, junk</td>
    <td class="tg-0pky">ASAP</td>
    <td class="tg-0pky">https://www.zooniverse.org/lab/14957 (if you can not access it, click "my projects" on the top left menu of zooniverse</td>
  </tr>
  <tr>
    <td class="tg-0pky">ERC WP1e/g</td>
    <td class="tg-0pky">"all"</td>
    <td class="tg-0pky">extract 350 vocs from CHI & FEM (vtc) for each recorded child, cut into 500ms (note: it would be good if the "skip/exclude" procedure would be in place)</td>
    <td class="tg-0pky">1) CHI/FEM/Junk; 2) if CHI or FEM, crying, laughing, canonical, non-canonical</td>
    <td class="tg-0pky">no rush</td>
    <td class="tg-0pky">doesn't exist yet</td>
  </tr>
  <tr>
    <td class="tg-0pky">zoo-phon (gold)</td>
    <td class="tg-0pky">1 pilot recording</td>
    <td class="tg-0pky">if there are enough, pull out 250 randomly & 250 from high child volubility 1-minute regions (the latter should be "consecutive vocs" ie take top minute and pull out all the vocs from there, then move to next minute, etc) - IS NOT chunkified!
</td>
    <td class="tg-0pky">NONE! will be annotated in the lab</td>
    <td class="tg-0pky">late Feb?</td>
    <td class="tg-0pky">(none)</td>
  </tr>
  <tr>
[...]

PR #130, opened 2021-02-12
Energy detection based sampler

lucasgautheron:
# Description

# Progress

 - [ ] tests
 - [x] documentation
 - [x] implementation
   - [x] **EnergyDetectionSampler** class 
   - [x] **energy computation** ~~with an optional lowpass filter. if the lowpass filter does not have any significant impact, we might as well just get rid of it, then we need no fft...~~ opted for an optional band-pass filter.
   - [x] ~~Handling saturation : this is especially important since targeting higher energies will favor saturated regions. We could discard regions where > 1 % of saturating signal. Though, what should we do in case there are > 1 channels ? discard the channels, and the whole segment if all channels are saturating ? or should we discard a segment when at least one of the channel is saturating ?~~ Marvin: let's keep it in mind, BUT no treatment for now
   - [ ] parallelise energy computation. the difficulty is to keep memory usage under control
   - [x] **filtering strategy**: threshold ? It cannot be hardcoded I think... Or sorting by highest energy and taking the N first windows ? Prolly too biased... For now I suggest to sample from the `x` percents of windows with the highest energy. What do you think ? Discussion [here](https://github.com/LAAC-LSCP/ChildRecordsData/pull/130#discussion_r575523023).
   - [x] **multi-channel support**: we need to decide how to do that in the first place !. I suggest we sum the energies across channels, and apply the threshold based on the quantile of the sum of energies. It does not make sense to sum amplitudes, but it does make much more sense to sum average energies over some interval of time. The only issue that I can see, is that this has a bias towards overlapping sources... but, so does any threshold based on the energy tbh ! Discussion [here](https://github.com/LAAC-LSCP/ChildRecordsData/pull/130#discussion_r575729473).
   - [ ] recordings filter : sample from a restricted list of recordings
   - [ ] **combining several recorders** : Let be x_i and y_i the fractions of windows having a higher energy than the window i, for the recorder x and the recorder y respectively. There are several possibilities:
     1. Check the agreement between devices ? (i.e. veto windows that do not pass the threshold in one of the devices). This leads to the condition (1-x_i > 1-q) and (1-y_i > 1-q).
     2. Find a way to combine the energy quantiles of each (window, recorder). These can be computed independently w/o any difficulty for the recorders x and y. By definition, P(x_i < x) = x and P(y_i < y) = y (they follow U(0,1)). It follows that P(x_i > x) = 1-x and P(y_i > y) = 1-y. Obviously, x_i and y_i are expected to be strongly correlated... But if we assume they are not, we might be able to combine these p-values in some way idk...
     3. We can also average energies or something, but that sounds really wrong... unless we do some normalization first, which is cumbersome...
     - Imo the agreement based approach seems more appropriate.
   
# Associated issues

#109


ChildProject/pipelines/sampler.py
```diff
@@ -79,6 +79,77 @@ def add_parser(samplers):
...
+
+    def compute_energy_loudness(self, chunk, sampling_frequency):
+        chunk_fft = fft([x/(2**15 - 1) for x in chunk.get_array_of_samples()])
+
+        if self.lowpass > 0:
+            chunk_fft = chunk_fft[:int(len(chunk)*self.lowpass/sampling_frequency)]
+        
+        return np.sum(np.abs(chunk_fft)**2)/len(chunk)
+
+    def get_recording_windows(self, profile, recording):
```
lucasgautheron, line 113: for now "profile" is ignored.

ChildProject/pipelines/sampler.py
```diff
@@ -79,6 +79,79 @@ def add_parser(samplers):
...
+
+        return windows
+        
+
+    def sample(self):
+        segments = []
+        for recording in self.project.recordings.to_dict(orient = 'records'):
+            windows = pd.DataFrame(self.get_recording_windows('', recording))
+            threshold = windows['energy'].quantile(0.1)
+            windows = windows[windows['energy'] >= threshold]
```
lucasgautheron, line 139: so, currently, we exclude the 10% of windows we the lowest energy, and sample from the rest. We could use much higher quantiles, e.g. 80% if we really want loudness. This avoids the use of arbitrary thresholds which is problematic for a number of reasons:
 - amplitudes are defined up to some factor, i.e. audio levels are relative, so how to define a universel threshold ?
 - it is difficult to make sense of some energy threshold, BUT it is much easier to make sense of a percentile based threshold. If we assume loud = speech, then, the threshold could be compared to the expected fraction of speech time of the recordings...

ChildProject/pipelines/sampler.py
```diff
@@ -79,6 +80,86 @@ def add_parser(samplers):
...
+        frequency = int(audio.frame_rate)
+
+        windows_starts = (1000*np.arange(self.windows_spacing, duration - self.windows_length, self.windows_spacing)).astype(int)
+        windows = []
+
+        for start in windows_starts:
+            energy = 0
+            chunk = audio[start:start+int(1000*self.windows_length)].get_array_of_samples()
+            
+            for channel in range(channels):
```
lucasgautheron, line 129: So, this is how we currently handle multiple channels : we sum the energy across channels.
It's always more meaningful than summing amplitudes... But is it really ?

ChildProject/pipelines/sampler.py
MarvinLvn: Agree with what has been said above ! 

If I follow you, the pipeline looks like this :
1) Exclude 80 % of windows w/ the lowest energies (which window length ? which time step (i.e. how much overlap between 2 successive windows ?))
2) Sample 10, 20 or N fixed-length windows amongst the 20% remaining windows.

Step 2) is important as we may want to be able to extract the same number of windows for 2 datasets whose cumulated duration is different (just using the quantile threshold doesn't allow us to do that).

For 2), I'd just use a uniform sampling : what do you think ? I think we can't skip the "try and listen" approach here. 
Don't hesitate to send me some extracted hunks :)

MarvinLvn:
About the multi-channel discussion :
- your approach seems fine to me.
- I think it'd be nice to be able to run the pipeline only on one of the N (N=2 or N=4) channel. In some cases (it may be the case with the BabyLogger), we can put some prior on which mic' gets the most speech.

MarvinLvn:
> combining several recorders : Let be x_i and y_i the fractions of windows having a higher energy than the window i, for the recorder x and the recorder y respectively. There are several possibilities:
> 1) Check the agreement between devices ? (i.e. veto windows that do not pass the threshold in one of the devices). This leads to the condition (1-x_i > 1-q) and (1-y_i > 1-q).
> 2) Find a way to combine the energy quantiles of each (window, recorder). These can be computed independently w/o any difficulty for the recorders x and y. By definition, P(x_i < x) = x and P(y_i < y) = y (they follow U(0,1)). It follows that P(x_i > x) = 1-x and P(y_i > y) = 1-y. Obviously, x_i and y_i are expected to be strongly correlated... But if we assume they are not, we might be able to combine these p-values in some way idk...
> 3) We can also average energies or something, but that sounds really wrong... unless we do some normalization first, which is cumbersome...Imo the agreement based approach seems more appropriate.

After your first quantile threshold, 3 possibilities :
1) No agreement at all amongst the 2 recorders : dead-end
2) You get agreement for some windows, but not enough of them 
3) Full agreement

The situation where you'll end up will mainly depend on the quantity of data you have, the threshold you apply.
In situation number 1) and 2), you could always decide to lower the quantile threshold until you get enough windows for recorder a) and recorder b) : not sure this is the ideal solution.

Solution that seems the easiest to implement is the one that consists in averaging energies.
You can do something like : 
1) Compute E_a, E_b the list of energies for recorder a) and recorder b)
2) Normalize E_a by : (E_a - mean(E_a)) / std(E_a), same for E_b
3) Go through your single-channel pipeline : apply quantile threshold, etc ...

This is just one more pre-processing step as compared to solution 1) and 2).
Agreement-based solutions seem scary to me as they are too data-dependent, and you'll always find situations where you won't have enough agreements between the 2 recorders. Hence, why I'll "combine the 2 recorders" as soon as possible.

lucasgautheron:
> About the multi-channel discussion :
> 
> * your approach seems fine to me.
> * I think it'd be nice to be able to run the pipeline only on one of the N (N=2 or N=4) channel. In some cases (it may be the case with the BabyLogger), we can put some prior on which mic' gets the most speech.

I suggest two different options:
 1. We provide an additional 'channel' option. If specified, only this channel is used to compute the energy
 2. We provide a 'channel_weights' option. It must be a list of length equal to the amount channels. Each channel energy is then weighted by some coefficient. One can turn off channels by setting their coefficient to 0.

lucasgautheron:
> > combining several recorders : Let be x_i and y_i the fractions of windows having a higher energy than the window i, for the recorder x and the recorder y respectively. There are several possibilities:
> > 
> > 1. Check the agreement between devices ? (i.e. veto windows that do not pass the threshold in one of the devices). This leads to the condition (1-x_i > 1-q) and (1-y_i > 1-q).
> > 2. Find a way to combine the energy quantiles of each (window, recorder). These can be computed independently w/o any difficulty for the recorders x and y. By definition, P(x_i < x) = x and P(y_i < y) = y (they follow U(0,1)). It follows that P(x_i > x) = 1-x and P(y_i > y) = 1-y. Obviously, x_i and y_i are expected to be strongly correlated... But if we assume they are not, we might be able to combine these p-values in some way idk...
> > 3. We can also average energies or something, but that sounds really wrong... unless we do some normalization first, which is cumbersome...Imo the agreement based approach seems more appropriate.
> 
> After your first quantile threshold, 3 possibilities :
> 
> 1. No agreement at all amongst the 2 recorders : dead-end
> 2. You get agreement for some windows, but not enough of them
> 3. Full agreement
> 
> The situation where you'll end up will mainly depend on the quantity of data you have, the threshold you apply.
> In situation number 1) and 2), you could always decide to lower the quantile threshold until you get enough windows for recorder a) and recorder b) : not sure this is the ideal solution.
> 
> Solution that seems the easiest to implement is the one that consists in averaging energies.
> You can do something like :
> 
> 1. Compute E_a, E_b the list of energies for recorder a) and recorder b)
> 2. Normalize E_a by : (E_a - mean(E_a)) / std(E_a), same for E_b
> 3. Go through your single-channel pipeline : apply quantile threshold, etc ...
> 
> This is just one more pre-processing step as compared to solution 1) and 2).
> Agreement-based solutions seem scary to me as they are too data-dependent, and you'll always find situations where you won't have enough agreements between the 2 recorders. Hence, why I'll "combine the 2 recorders" as soon as possible.

I see.. Though, I am more optimistic than you. I would expect high correlations between the two recorders. Besides the distribution of energies is very flat. Here is the distribution for 600 30s windows drawn from a 20h USB recording. The energy is spanned across 5 orders of magnitude.

Screenshot 2021-02-17 at 11 58 17

In the end, it is not very difficult to switch from one way or another once we have the core pipeline ready.
So from now on, i'll implement the method you suggest, but the recorder-agreement thing will lie outside of the package, because that seems way too specific - though it will use the package to compute the energies.

There is just one thing I need to think about : whether this z-score normalization really fits. I might be overthinking this...

ChildProject/pipelines/sampler.py
lucasgautheron: > Agree with what has been said above !
> 
> If I follow you, the pipeline looks like this :
> 
> 1. Exclude 80 % of windows w/ the lowest energies (which window length ? which time step (i.e. how much overlap between 2 successive windows ? are those hard-coded ?))
> 2. Sample 10, 20 or N fixed-length windows amongst the 20% remaining windows.
> 
> Step 2) is important as we may want to be able to extract the same number of windows for 2 datasets whose cumulated duration is different (just using the quantile threshold doesn't allow us to do that).

You are right. Regarding the parameters, they are options:

```bash
$ child-project sampler . energy-detection --help
usage: child-project sampler path energy-detection [-h] --windows-length
                                                   WINDOWS_LENGTH
                                                   --windows-spacing
                                                   WINDOWS_SPACING
                                                   --windows-count
                                                   WINDOWS_COUNT
                                                   [--threshold THRESHOLD]
                                                   [--low-freq LOW_FREQ]
                                                   [--high-freq HIGH_FREQ]

optional arguments:
  -h, --help            show this help message and exit
  --windows-length WINDOWS_LENGTH
                        length of each window (in seconds)
  --windows-spacing WINDOWS_SPACING
                        spacing between windows (in seconds)
  --windows-count WINDOWS_COUNT
                        how many windows so sample from
  --threshold THRESHOLD
                        lowest energy quantile to sample from. default is 0.8
                        (i.e., sample from the 20% windows with the highest
                        energy).
  --low-freq LOW_FREQ   remove all frequencies below low-freq before
                        calculating each window's energy. (in Hz)
  --high-freq HIGH_FREQ
                        remove all frequencies above high-freq before
                        calculating each window's energy. (in Hz)
```

e.g.:

```bash
child-project sampler examples/valid_raw_data/ energy-detection \ 
    --windows-length 30 \
    --windows-spacing 120 \
    --windows-count 10 \
     --low-freq 100 \
     --high-freq 4000 \
     --threshold 0.9 \
     out.csv
```
 
> For 2), I'd just use a uniform sampling : what do you think ? I think we can't skip the "try and listen" approach here.
> Don't hesitate to send me some extracted hunks :)
> 
> EDIT : 2) is for the general case (when we don't want to compare between 2 recorders)

2 is indeed a uniform sampling. 
I will extract chunks. Here is what I can do : extract 5 chunks randomly for each energy decile. Is that good for you ?

lucasgautheron: This will be very slow if you use it on a whole dataset...

```python
import os
import pandas as pd
from pydub import AudioSegment

from ChildProject.projects import ChildProject
from ChildProject.pipelines.sampler import EnergyDetectionSampler

project = ChildProject("examples/valid_raw_data")
project.read()

sampler = EnergyDetectionSampler(project, 30, 120, 500, 0)
sampler.sample()
sampler.segments.sort_values('energy', inplace = True)
sampler.segments['pos'] = 1
sampler.segments['decile'] = (10*(sampler.segments['pos'].cumsum()-1)/sampler.segments.shape[0]).astype(int)

chunks = sampler.segments.groupby('decile').sample(5)

for chunk in chunks.to_dict(orient = 'records'):
    recording_path = os.path.join(project.path, ChildProject.RAW_RECORDINGS, chunk['recording_filename'])
    audio = AudioSegment.from_wav(recording_path)

    onset = int(chunk['segment_onset']*1000)
    offset = int(chunk['segment_offset']*1000)

    print(os.path.join('windows', str(chunk['decile']), '{}_{}.mp3'.format(onset, offset)))

    os.makedirs(os.path.join('windows', str(chunk['decile'])), exist_ok = True)
    audio[onset:offset].export(os.path.join('windows', str(chunk['decile']), '{}_{}.mp3'.format(onset, offset)), format = 'mp3')
```

alecristia: not sure if this will be helpful, but there's a slither of data human-coded from children who wore the previous babylogger+LENA (at the same time). It sounds like these data could provide a human gold standard. I don't know where the annotations or audios are -- Xuan Nga oversaw them, and they were used for the pygamma examples in the seshat paper.

lucasgautheron: Manel promised me some BabyLogger + LENA data, I guess that was it ? I'll ask her!

PR #176, opened 2021-03-21
.eaf builder

lucasgautheron:
- [x] tests
Use a periodic sampler on examples/valid_raw_data/recordings/raw/sound.wav than one of the template at random, parse the template with pympi and check code_, context_ and code_num tiers
- [x] documentation
    - [x] code
    - [x] tools section 
- [x] implementation
    - [x] core
    - [x] custom templates


ChildProject/pipelines/eafbuilder.py
```diff
@@ -1,28 +1,120 @@
...
+        :param template: name of the template to use (basic, native, or non-native)
+        :type template: str
+        :param context_onset: context onset and segment offset difference in milliseconds, 0 for no introductory context
+        :type context_onset: float
+        :param context_offset: context offset and segment offset difference in milliseconds, 0 for no outro context
+        :type context_offset: float
+        """
+
+        # TODO: Make sure etf file paths are approprite and robust. 
+        etf_path = "{}.etf".format(template)
```
alecristia, line 81: these are not always provided, right? why do we want them?

ChildProject/pipelines/eafbuilder.py
```diff
@@ -1,28 +1,120 @@
...
+
+        print("making the "+eaf_type+" eaf file and csv")
+
+        segments = pd.read_csv(segments)
+
+        for recording_filename, segs in segments.groupby('recording_filename'):
+            recording_filename = os.path.splitext(recording_filename)[0]
+            output_filename = recording_filename + '_' + eaf_type + '_' + template
+            
+            # TODO: This list of timestamps as tuples might not be ideal/should perhaps be optimized, but I am just replicating the original eaf creation code here.
```
alecristia, line 92: what is the alternative -- is it a csv file with on/off?

ChildProject/pipelines/eafbuilder.py
```diff
@@ -1,28 +1,120 @@
...
-        parser.add_argument("path", help = "project path")
-        parser.add_argument("destination", help = "eaf destination")
+        parser.add_argument("--destination", help = "eaf destination")
         parser.add_argument('--segments', help = 'path to the input segments dataframe', required = True)
-        parser.add_argument('--type', help = 'eaf-type', choices = ['random', 'periodic'])
-        parser.add_argument('--template', help = 'eaf template', choices = ['basic', 'native', 'non-native'])
-        parser.add_argument('--context-onset', help = 'context onset and segment offset difference, 0 for no introductory context', type = float, default = 0)
-        parser.add_argument('--context-offset', help = 'context offset and segment offset difference, 0 for no outro context', type = float, default = 0)
+        # TODO: add other options here such as high-volubility, energy, etc.?
+        parser.add_argument('--eaf-type', help = 'eaf-type', choices = ['random', 'periodic'], required = True)
```
alecristia, line 117: agreed! not totally sure about the energy one being used a lot in the future, but high vol is very common (as you saw in yesterday's talk)

ChildProject/pipelines/samplers.py
```diff
@@ -357,6 +357,10 @@ def _segment_scores(self, recording):
         segments = self.retrieve_segments(recording['recording_filename'])
 
         if segments is None:
+            print("warning: no annotations from the set '{}' were found for the recording '{}'".format(
+                self.annotation_set,
+                recording['recording_filename']
+            ))
             return pd.DataFrame(columns = ['segment_onset', 'segment_offset', 'recording_filename'])
 
         # NOTE: The timestamps were simply rounded to 5 minutes in the original
```
alecristia, line 366: what is the alternative here?
I should mention that LENA users typically have time-stamped audios that make reference to local time, so they'll use 5 min chunks of clock time; VTC users may not always have this (at least right now), so we tend to do time from the onset of the file.

ChildProject/pipelines/eafbuilder.py
lucasgautheron: Right now we provide 3 templates with our package (so the users don't need their own), and only these can be used.

However, we could allow users to use their own, custom templates as well.

Or, we could remove the default templates from our package, and oblige users to manager their own templates.

What do you think is better ?

sarpu: To chime in, these templates were provided with the original eaf builder code

ChildProject/pipelines/eafbuilder.py
sarpu: Yeah, essentially that is what the eaf file comes to be. Just a bunch of tiers with empty annotations with the segments. So we can output csv too, but that is already output by the samplers. Though I placed that TODO more in order to double check if this is the efficient way to pull out segments given a pandas dataframe than anything else.

lucasgautheron: I think the comment is just about the efficiency of the code, not about its behavior right @sarpu ?

so nothing to do with the input :)

ChildProject/pipelines/eafbuilder.py
lucasgautheron: so maybe we should allow arbitrary values for this parameter ?

ChildProject/pipelines/samplers.py
lucasgautheron: Just to clarify: 

We have made the length of the clips a parameter, so any value can be used (30 s, 1 minute, 5 minutes, etc.)

But I think we should add offset parameter to all window-based samplers, so that people can align them as they wish (e.g. beginning at a specific clocktime). This is also useful when you need to align windows for simultaneous recordings from different devices.

I've done that for the PeriodicSampler and the EnergySampler.

ChildProject/pipelines/eafbuilder.py
sarpu: @lucasgautheron Yep, I just place my TODOs a bit too liberally

PR #186, opened 2021-04-20
detect and load additional metadata recursively

lucasgautheron:
### Description

Currently, recordings and children metadata are assumed to be entirely stored in metadata/recordings.csv and metadata/children.csv respectively. However, in case users spread their metadata across several files (e.g. to achieve multi-tier access), then there is no standard.

I suggest the standard should be that additional metadata belong as extra CSVs in metadata/recordings and metadata/children respectively. These folders are optional.

If the package detects them, it will list all CSVs from these folders **recursively**,  sort them alphabetically (currently based on their basename, which might be ambiguous), then merge them sequentially.

For instance, if recordings/0.csv and recordings/1.csv exist,
the package will :

 - load metadata/recordings.csv as a dataframe
 - load recordings/0.csv, merge it with the previous dataframe; in case of duplicate columns, recordings/0.csv will prevail
 - load recordings/1.csv, merge it with the previous dataframe; in case of duplicate columns, recordings/1.csv will prevail

etc.

### Status

- [x] tests
- [x] documentation
- [ ] implementation 
    - [x] merge additional metadata recursively (alphabetical order)
    - [ ] improve tests so that they can identify problematic files... this is a big change.

### DIFF
--- a/docs/source/format.rst
+++ b/docs/source/format.rst
@@ -57,22 +57,52 @@ the standards detailed right below.
 Metadata
 --------
 
-children notebook
+Children notebook
 ~~~~~~~~~~~~~~~~~
 
-The children dataframe needs to be saved at ``metadata/children.csv``.
+The children metadata dataframe needs to be saved at ``metadata/children.csv``.
 
 .. index-table:: Children metadata
    :header: children
 
-recording notebook
-~~~~~~~~~~~~~~~~~~
 
-The recordings dataframe needs to be saved at
+Recordings notebook
+~~~~~~~~~~~~~~~~~~~
+
+The recordings metadata dataframe needs to be saved at
 ``metadata/recordings.csv``.
 
 .. index-table:: Recordings metadata
    :header: recordings
+
+Splitting the metadata in several files
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Sometimes, access to parts of the metadata should be limited
+to a list of authorized users. This can be achieved by moving confidential
+information out of the main notebook to a separate CSV file to
+be only delivered to authorized users. These additional files
+should be placed according to the table below:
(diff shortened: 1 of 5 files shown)


docs/source/format.rst
```diff
@@ -57,23 +57,53 @@ the standards detailed right below.
...
+
+
+.. csv-table:: Additional metadata
+   :header: data,main notebook,location of additional notebooks
+
+   children,``metadata/children.csv``,``metadata/children/``
+   recordings,``metadata/recordings.csv``,``metadata/recordings/``
+
+There can be as many additional notebooks as necessary, and recursion
+if permitted.
```
alecristia, line 95: if permitted --> is permitted

docs/source/format.rst
```diff
@@ -57,23 +57,53 @@ the standards detailed right below.
...
+
+.. note::
+
+   In case two or more notebooks contain the same column, the files
+   whose names come first in alphabetical order will prevail while
+   loading the dataset with our package. For instance, if
+   ``child_dob`` is specified in both  ``metadata/recordings/0_private.csv``
+   and ``metadata/recordings/1_public.csv``, the values in the former file will prevail if it is available.
+   This is useful when anonymized values for a certain parameter still need to be shared,
+   but should replaced with the true values for those who have access to the full dataset.
```
alecristia, line 105: but should *be* replaced

alecristia:
> sort them alphabetically (currently based on their basename, which might be ambiguous)

I didn't see in your code a check for ambiguity (ie without path, the names are identical).

I may have also missed a set of messages for the user:

- multi-tier detected
- merging x and y (in that order of precedence)

and/or error:

- basenames are identical -- please disambiguate precedence for multi-tier metadata

(my messages are super cryptic, but you see the idea!)

PR #228, opened 2021-06-04
multiple storage support for annotations: add gzip

lucasgautheron:
(proof of concept)


lucasgautheron:
According to [this benchmark](https://www.architecture-performance.fr/ap_blog/loading-data-into-a-pandas-dataframe-a-performance-study/), it is not worth using HDF_Fixed in any case.

So we should use HDF_Table, taking advantage of categoricals, AND maybe merging annotations from different recordings together (which will require a bit of work and will make it harder to use datasets without the python API)

PR #254, opened 2021-07-10
LENA/ACLEW metrics pipeline

lucasgautheron:
- [x] tests
- [x] documentation
- [x] implementation
    - [x] LENA
    - [x] ACLEW


tests/truth/aclew_metrics.csv
```diff
@@ -0,0 +1,2 @@
+child_id,voc_fem_ph,voc_dur_fem_ph,avg_voc_dur_fem,voc_chi_ph,voc_dur_chi_ph,avg_voc_dur_chi,wc_fem_ph,sc_fem_ph,pc_fem_ph,wc_adu_ph,sc_adu_ph,pc_adu_ph,cry_voc_chi_ph,cry_voc_dur_chi_ph,can_voc_chi_ph,can_voc_dur_chi_ph,avg_can_voc_dur_chi,non_can_voc_chi_ph,non_can_voc_dur_chi_ph,avg_non_can_voc_dur_chi,lp_n,cp_n,lp_dur,cp_dur,duration
+1,3600.0,360.0000000000002,0.10000000000000006,9000.0,900.0,0.1,3600.0,7200.0,14400.0,3600.0,7200.0,14400.0,0,0,4500.0,449.99999999999994,0.09999999999999999,4500.0,450.0,0.1,1.0,0.5,1.0,0.49999999999999994,8.0
```
alecristia, line 2: are the numbers with many decimals part of the test?

ChildProject/pipelines/metrics.py
```diff
@@ -0,0 +1,391 @@
...
+            self.recordings = recordings['recording_filename'].tolist()
+        elif isinstance(recordings, pd.Series):
+            self.recordings = recordings.tolist()
+        elif isinstance(recordings, list):
+            self.recordings = recordings
+        else:
+            if not os.path.exists(recordings):
+                raise ValueError(
+                    "'recordings' is neither a pandas dataframe,"
+                    "nor a list or a path to an existing dataframe."
```
alecristia, line 51: can we give the user an instruction RE what to do about this?

alecristia: eg check parameters passed to the function, or check structure of dataframe against instructions LINK

ChildProject/pipelines/metrics.py
```diff
@@ -0,0 +1,391 @@
...
+        alice: str = 'alice',
+        vcm: str = 'vcm',
+        recordings: Union[str, List[str], pd.DataFrame] = None,
+        by: str = 'recording_filename',
+        threads: int = 1):
+
+        super().__init__(project, by, recordings)
+
+        self.vtc = vtc
+        self.alice = alice
```
alecristia, line 233: have you checked what happens if the user doesn't have an alice analysis? or (this won't typically happen) a vtc analysis?

alecristia: (or a vcm analysis -- which does happen quite often outside of ACLEW data)

alecristia (commented):
i only left a few questions!

tests/truth/aclew_metrics.csv
lucasgautheron: The truth was generated, so there are some rounding errors, but pandas testing function have some tolerance so it works anyway!

ChildProject/pipelines/metrics.py
lucasgautheron: It will work if VCM and/or ALICE are missing, but the VTC is required.
I could mention it in the docs.
Do you think the VTC should be optional?

alecristia: rather, I think if VCM or ALICE are missing, it would be helpful to give a warning that tells the user that we're not computing those bc they are missing (and if they think they are not missing, to check parameters).
For VTC, instead, it should be an error saying that vtc is not found, and check parameters if they do have it.

lucasgautheron: Done 30f4236

PR #266, opened 2021-08-20
get annotations within a certain time range

lucasgautheron:
- [x] tests
    - [x] `get_within_time_range`
    - [x] `get_segments_timestamps`
- [x] documentation
- [x] implementation
    - [x] `get_within_time_range`
    - [x] `get_segments_timestamps`
    - [ ] CLI

### DIFF
--- a/ChildProject/annotations.py
+++ b/ChildProject/annotations.py
@@ -1011,7 +1011,7 @@ class AnnotationManager:
         return segments
 
     def get_within_time_range(
-        self, annotations: pd.DataFrame, start_time: str, end_time: str
+        self, annotations: pd.DataFrame, start_time: str, end_time: str, errors="raise"
     ):
         """Clip all input annotations within a given HH:MM clock-time range.
         Those that do not intersect the input time range at all are filtered out.
@@ -1023,6 +1023,8 @@ class AnnotationManager:
         :type start: str
         :param end: offset HH:MM clocktime
         :type end: str
+        :param errors: how to deal with invalid start_time values for the recordings
+        :type errors: str
         :return: a DataFrame of annotations;
         For each row, ``range_onset`` and ``range_offset`` are clipped within the desired clock-time range.
         The clock-time corresponding to the onset and offset of each annotation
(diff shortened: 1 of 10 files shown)


ChildProject/metrics.py
```diff
@@ -244,7 +244,7 @@ def gamma(
         )
 
     dissim = CombinedCategoricalDissimilarity(
-        list(continuum.categories), delta_empty=1, alpha=alpha, beta=beta
+        delta_empty=1, alpha=alpha, beta=beta
```
alecristia, line 247: this is unrelated, right?

tests/test_annotations.py
```diff
@@ -11,310 +11,462 @@
...
+    annotations = pd.read_csv("tests/data/time_range_annotations.csv")
+    matches = am.get_within_time_range(annotations, '09:00', '20:00')
+
+    truth = pd.read_csv('tests/truth/time_range.csv')
+
+    pd.testing.assert_frame_equal(
+        standardize_dataframe(matches, truth.columns),
+        standardize_dataframe(truth, truth.columns)
+    )
+
```
alecristia, line 269: maybe i'm missing something, but do you have a check for the user inputting start/stop with wrong formatting? (eg 1pm or 1.13 etc)

alecristia (approved):

ChildProject/metrics.py
lucasgautheron: yes; the tests were failing due to an upgrade of pygamma-agreement. had to change this for the tests to pass again

tests/test_annotations.py
lucasgautheron: Added (a7c60cb4c8b14b13221645c89709f96da329c75e)

ChildProject/annotations.py
```diff
@@ -1034,8 +1036,19 @@ def get_within_time_range(
...
+            raise ValueError(
+                f"invalid value for start_time ('{start_time}'); should have HH:MM format instead"
+            )
+
+        try:
+            end_dt = datetime.datetime.strptime(end_time, "%H:%M")
+        except:
+            raise ValueError(
+                f"invalid value for end_time ('{end_time}'); should have HH:MM format instead"
+            )
```
alecristia, line 1052: we should foresee one more error category: start time is after end time

tests/test_annotations.py
```diff
@@ -267,6 +267,14 @@ def test_within_time_range(project):
...
     )
 
+    exception_caught = False
+    try:
+       matches = am.get_within_time_range(annotations, '9am', '8pm')
+    except ValueError as e:
+        exception_caught = True
+
+    assert exception_caught, "no exception was thrown despite invalid times"
+
```
alecristia, line 277: that's kind :)

alecristia (approved):

lucasgautheron:
@alecristia, do you think any of these two functionalities are worth being implemented into the command-line interface?

alecristia (approved):
I don't think this needs to be in the CLI -- the python API is all we need!

PR #272, opened 2021-09-05
Time-unit aggregated metrics

lucasgautheron:
- [x] tests
- [x] documentation
- [ ] implementation
    - [ ] period_origin
 - easy to that with `pandas.grouper`
 - more difficult with `pandas.date_range`
    - [x] handling of periods only partially annotated (this is important...). Also, handling of periods covered by several annotations, i.e. when several recordings are aggregated (--by child_id) (WIP)
    - [ ] update the `ItsConverter` to import each segment timestamp as well
    - [ ] handling of segments overlapping several periods? (is it worth it?)
    - [x] handle recordings longer than 24h (low-priority)


lucasgautheron:
Another thing that we may want to take care of: LENA recordings can have pauses; in this case, calculating `onset_time` from `start_time` and `segment_onset` will lead to incorrect values; it is necessary, in such cases, to rely on LENA values rather than values calculated from the annotations+metadata. I suggest that the LENA converter should import the timestamps too, and that users should be able to use imported timestamps instead. What do you think @alecristia?

ChildProject/pipelines/metrics.py
```diff
@@ -472,6 +469,156 @@ def add_parser(subparsers, subcommand):
...
+            closed="left",
+        )
+
+    def _process_unit(self, unit: str):
+        annotations, segments = self.retrieve_segments([self.set], unit)
+
+        segments = self.am.get_segments_timestamps(segments, ignore_date=True)
+        segments.dropna(subset=["onset_time"], inplace=True)
+
+        segments["onset_time"] -= pd.to_timedelta(
```
alecristia, line 516: hmm i'm lost here, it would be great to have a comment explaining what this section does...

alecristia (approved):
i'm not sure i've followed the logic throughout -- perhaps when i see the doc it'll be obvious, but otherwise, a comment on the scope would be useful. For instance, can unit be of any size? or is it hourly? one section divides by the number of sections in a day... ?

alecristia:
> Another thing that we may want to take care of: LENA recordings can have pauses; in this case, calculating `onset_time` from `start_time` and `segment_onset` will lead to incorrect values; it is necessary, in such cases, to rely on LENA values rather than values calculated from the annotations+metadata. I suggest that the LENA converter should import the timestamps too, and that users should be able to use imported timestamps instead. What do you think @alecristia?

Yes, I thought when there are pauses in LENA recs, we split up the long .wav, and then impute the onset times of each non-pause using information from the LENA .its. Is that not how things are working right now?

ChildProject/pipelines/metrics.py
lucasgautheron: > > Another thing that we may want to take care of: LENA recordings can have pauses; in this case, calculating `onset_time` from `start_time` and `segment_onset` will lead to incorrect values; it is necessary, in such cases, to rely on LENA values rather than values calculated from the annotations+metadata. I suggest that the LENA converter should import the timestamps too, and that users should be able to use imported timestamps instead. What do you think @alecristia?
> 
> Yes, I thought when there are pauses in LENA recs, we split up the long .wav, and then impute the onset times of each non-pause using information from the LENA .its. Is that not how things are working right now?

We do that for EL1000, but it is possible that some people decide to do otherwise.

lucasgautheron: > hmm i'm lost here, it would be great to have a comment explaining what this section does...
Sure, sorry. I just added more comments, does it help?

lucasgautheron:
> i'm not sure i've followed the logic throughout -- perhaps when i see the doc it'll be obvious, but otherwise, a comment on the scope would be useful. For instance, can unit be of any size? or is it hourly? one section divides by the number of sections in a day... ?

I hope the comments  I just added help.
Also you are right that I did not document things yet;

Days are divided according to the period chosen by the user (for instance: 10 minutes, 30 minutes, 1 hour, 4 hours, ... it is entirely up to the user)

ChildProject/pipelines/metrics.py
alecristia: RE allowing others to choose not to split the recording, let's not! I think it'll make data more heterogeneous in a non-interesting way.
I believe the reason why you had suggested not enforcing it had to do with recordings where there are uncertainties as to onset time, as in Vanuatu -- but that never happens with LENA recordings.
So I think it's safe to enforce it for LENA.
Would you like us to think about it some more in the case of non-LENA recs with timing uncertainty?

ChildProject/pipelines/metrics.py
```diff
@@ -510,9 +510,14 @@ def __init__(
...
         annotations, segments = self.retrieve_segments([self.set], unit)
 
+        # retrieve timestamps for each vocalization, ignoring the day of occurence
         segments = self.am.get_segments_timestamps(segments, ignore_date=True)
+
+        # dropping segments for which no time information is available
         segments.dropna(subset=["onset_time"], inplace=True)
 
+        # update the timestamps so that all vocalizations appear
+        # to happen on the same day
```
alecristia, line 520: does that mean --- put them all in a single time line, consecutive, so that the user gets numbers every eg 5 mins to the end of the rec?

lucasgautheron: this really is just a technicality to allow pandas.Grouper to group together all vocs that happen between e.g. 08:00 and 09:00 (in case of 1-hour periods) regardless of the day they happened. This is useful for recordings lastin > 24h (which do not exist yet), but much more importantly, this is required for recordings that end after midnight.

ChildProject/pipelines/metrics.py
lucasgautheron: This was only one of the arguments for not enforcing, the other being that this makes it harder to convert LENA datasets and to maintain them if all annotations are based on the whole, "unsplit" recordings.
But it sounds reasonable to decide, at least for some time or until someone else offers to do it, to restrict this pipeline to "contiguous" recordings.

To answer your last question: for now, I think the package should assume that the recordings are contiguous. This means that Vanuatu's case could be treated either of two ways:

 1. Using the package pipeline upstream, and then accounting for the uncertainty in the time bins in the model.
 2. Or maybe developing a different pipeline specific to this case.

lucasgautheron:
This is what the output looks like (some rows are incorrect, but this is expected at this point -- the implementation is not finished for recordings that overlap midnight)

https://gist.github.com/lucasgautheron/0edf59f8436f0247fb3c61a194bb3e14

PR #274, opened 2021-09-07
--recordings option for all audio processors

lucasgautheron:
Introducing a new option to process only specific recordings.

- [x] documentation
- [x] tests
- [x] implementation


lucasgautheron:
@alecristia, would you like to test this new feature on Vanuatu?

PR #280, opened 2021-09-11
compute-durations update

lucasgautheron:
compute-durations should be usable even when not all recordings are locally available.


lucasgautheron:
I think this may only work properly if the metadata has no 'NA' recording_filename, so this is something we might want to enforce before merging.

PR #284, opened 2021-09-15
validate custom recording profiles or annotation sets

lucasgautheron:
- [x] tests
- [x] documentation
- [x] implementation
    - [x] --profile option
    - [x] --annotations option
    - [x] support paths in the arguments


lucasgautheron:
Just realized the wildcard is a bad idea, because it will be interpreted and expanded by bash... -> replacement needed!

alecristia (commented):
much clearer!

PR #328, opened 2021-11-23
Conversation sampler

lucasgautheron:
- [x] implementation
- [x] docs
- [x] tests

### DIFF
--- a/ChildProject/pipelines/samplers.py
+++ b/ChildProject/pipelines/samplers.py
@@ -641,6 +641,12 @@ class HighVolubilitySampler(Sampler):
     with the highest volubility from each recording,
     as calculated from the metric ``metric``.
 
+    ``metrics`` can be any of three values: words, turns, and vocs.
+
+     - The **words** metric sums the amount of words within each window. For LENA annotations, it is equivalent to **awc**.
+     - The **turns** metric (aka ctc) sums conversational turns within each window. It relies on **lena_conv_turn_type** for LENA annotations. For other annotations, turns are estimated as adult/child speech switches in close temporal proximity.
+     - The **vocs** metric sums vocalizations within each window. If ``metric="vocs"`` and ``speakers=['CHI']``, it is equivalent to the usual cvc metric (child vocalization counts).
+
     :param project: ChildProject instance of the target dataset.
     :type project: ChildProject.projects.ChildProject
     :param annotation_set: set of annotations to calculate volubility from.
@@ -720,7 +726,6 @@ class HighVolubilitySampler(Sampler):
                     ["TIFR", "TIMR"]
                 )
             else:
-                segments.sort_values(["segment_onset", "segment_offset"])
                 segments = segments[segments["speaker_type"].isin(self.speakers)]
                 segments["iti"] = segments["segment_onset"] - segments[
                     "segment_offset"
--- a/ChildProject/pipelines/samplers.py
+++ b/ChildProject/pipelines/samplers.py
@@ -978,7 +978,7 @@ class ConversationSampler(Sampler):
         )
         parser.add_argument(
             "--interval",
-            help="maximum time-interval between two vocalizations (in milliseconds) to consider them to be part of the same conversational block. default is 1000",
+            help="maximum time-interval between two consecutive vocalizations (in milliseconds) to consider them to be part of the same conversational block. default is 1000",
             default=1000,
             type=int,
         )
--- a/ChildProject/pipelines/samplers.py
+++ b/ChildProject/pipelines/samplers.py
@@ -850,6 +850,28 @@ class HighVolubilitySampler(Sampler):
 
 
 class ConversationSampler(Sampler):
+    """Conversation sampler
+
+    :param project: ChildProject instance
+    :type project: ChildProject.projects.ChildProject
+    :param annotation_set: set of annotation to derive conversations from
+    :type annotation_set: str
+    :param count: amount of conversations to sample
+    :type count: int
+    :param interval: maximum time-interval between two consecutive vocalizations (in milliseconds) to consider them part of the same conversational block, defaults to 1000
+    :type interval: int, optional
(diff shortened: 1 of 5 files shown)


lucasgautheron:
@alecristia I made the High-vol sampler more "generic" by replace 'cvc', 'awc' and 'ctc' with 'vocs', 'words' and 'turns'.
Basically you can do as much as before but you get to choose the speakers to target for each metric.
Is this fine with you?

alecristia:
> @alecristia I made the High-vol sampler more "generic" by replace 'cvc', 'awc' and 'ctc' with 'vocs', 'words' and 'turns'. Basically you can do as much as before but you get to choose the speakers to target for each metric. Is this fine with you?

YES! let me take a quick look...

ChildProject/pipelines/samplers.py
```diff
@@ -645,7 +645,7 @@ class HighVolubilitySampler(Sampler):
     :type project: ChildProject.projects.ChildProject
     :param annotation_set: set of annotations to calculate volubility from.
     :type annotation_set: str
-    :param metric: the metric to evaluate high-volubility. should be any of 'awc', 'ctc', 'cvc'.
+    :param metric: the metric to evaluate high-volubility. should be any of 'words', 'turns', 'vocs'.
```
alecristia, line 648: hmm reading it like this I think LENA users will be confused. Perhaps words (awc), turns (ctc), vocs (cvc)?

ChildProject/pipelines/samplers.py
```diff
@@ -711,74 +713,65 @@ def _segment_scores(self, recording):
...
             else:
                 segments.sort_values(["segment_onset", "segment_offset"])
-                segments = segments[
-                    segments["speaker_type"].isin(["FEM", "MAL", "CHI"])
-                ]
-                segments["duration"] = (
-                    segments["segment_offset"] - segments["segment_onset"]
-                )
+                segments = segments[segments["speaker_type"].isin(self.speakers)]
                 segments["iti"] = segments["segment_onset"] - segments[
```
alecristia, line 725: Right, so as to leave a note on the logic:
- for LENA, we just follow LENA's counts
- for ACLEW, we select turns concerning the speakers of choice

ChildProject/pipelines/samplers.py
```diff
@@ -711,74 +713,65 @@ def _segment_scores(self, recording):
...
+                segments["is_CT"] = (segments["iti"] < 1000) & (
+                    (
+                        (segments["speaker_type"] == "CHI")
+                        & (segments["prev_speaker_type"].isin(key_child_environment))
+                    )
+                    | (
+                        (segments["speaker_type"].isin(key_child_environment))
+                        & (segments["prev_speaker_type"] == "CHI")
+                    )
                 )
```
alecristia, line 741: simplification of lambda/axis - no negative/different behavior consequences that you can think of, right?

alecristia: (oops, that was intended as a comment, not a review item)

ChildProject/pipelines/samplers.py
```diff
@@ -711,74 +713,65 @@ def _segment_scores(self, recording):
...
             else:
                 segments = (
-                    segments[segments.speaker_type == "CHI"]
+                    segments[segments.speaker_type.isin(self.speakers)]
                     .groupby("chunk", as_index=False)[["segment_onset"]]
                     .count()
-                    .rename(columns={"segment_onset": "cvc"})
+                    .rename(columns={"segment_onset": self.metric})
                     .merge(windows)
                 )
```
alecristia, line 768: very elegant!

docs/source/samplers.rst
```diff
@@ -86,3 +86,22 @@ High-Volubility sampler
...
 
    child-project sampler /path/to/dataset /path/to/destination high-volubility --help
+
+Conversation sampler
+~~~~~~~~~~~~~~~~~~~~
+
+The conversation sampler returns the conversational blocks with the highest amount of turns.
+The first step is the detection of conversational blocks.
+Two consecutive vocalizations are considered part of the same conversational block if they are not separated
+by an interval longer than a certain duration, which by default is set to 1000 milliseconds.
```
alecristia, line 96: should we perhaps clarify that this is only for the ACLEW pipeline? if I read the code correctly above, the filter for 1k ms is not applied to subset from the TIFR/TIMR

docs/source/samplers.rst
```diff
@@ -86,3 +86,22 @@ High-Volubility sampler
...
+
+Then, the amount of conversational turns (between the key child and adult) is calculated for each conversational block.
+The sampler returns, for each unit, the desired amount of conversations with the higher amount of turns.
```
alecristia, line 99: ah, I read the code above as suggesting that I can select only FEM-CHI blocks -- did I get that wrong?

ChildProject/pipelines/samplers.py
lucasgautheron: nope, if anything it should be faster!

docs/source/samplers.rst
lucasgautheron: done!

docs/source/samplers.rst
lucasgautheron: you can limit the search to FEM vs CHI only, but by default the search includes all both FEM and MAL. It is more clear now ?

ChildProject/pipelines/samplers.py
lucasgautheron: Now mentioned in the docs

ChildProject/pipelines/samplers.py
lucasgautheron: I tried to make it more clear in the docs (both the API and the cli)

docs/source/samplers.rst
alecristia: Then, the amount of conversational turns (between the key child and adult) --> Then, the amount of conversational turns (by default, between the key child and female/male adult)

PR #334, opened 2022-01-10
build pre-filled eaf based on previous diarization annotations

lucasgautheron:
- [x] tests
- [x] documentation
- [x] implementation


ChildProject/pipelines/eafbuilder.py
```diff
@@ -51,6 +53,37 @@ def create_eaf(
...
+            if "speaker_id" in segment:
+                speaker_id = segment["speaker_id"]
+            elif "speaker_type" in segment and segment["speaker_type"] in type_to_id:
+                speaker_id = type_to_id[segment["speaker_type"]]
+
+            if speaker_id is None:
+                continue
+
+            if speaker_id not in eaf.tiers:
+                eaf.add_tier(speaker_id)
```
lucasgautheron, line 79: don't know if we should set other parameters of add_tier and if that case, which values should be assigned. e.g. what about ling? See https://github.com/dopefishh/pympi/blob/a41807e0b7d20fb014b95bc836a366933e7fe4a2/pympi/Elan.py#L375-L395

ChildProject/pipelines/eafbuilder.py
```diff
@@ -51,6 +53,37 @@ def create_eaf(
...
+            val: key
+            for key, val in reversed(
+                list(AnnotationConverter.SPEAKER_ID_TO_TYPE.items())
+            )
+        }
+
+        for segment in speech_segments.to_dict(orient="records"):
+            speaker_id = None
+
+            if "speaker_id" in segment:
```
lucasgautheron, line 70: This won't work for the LENA, because LENA defines speaker_id, but the values are not commensurable to that of our eaf templates.

lucasgautheron:
Example:

```bash
child-project eaf-builder --destination ../test/eaf --segments ../test/segments_20220110_174937.csv --eaf-type random --template basic --path . --import-speech-from textgrid/mm
```

lucasgautheron:
FYI, the only thing that prevents merging at this step is the lack of test coverage for this feature! Will do it ASAP.

LoannPeurey:
additional modifications :
- prefix imported tiers with the name of the set (eg set vtc for speaker CHI => tier named VTC-CHI
- replace NA values for speaker_type by SPEECH for some specific formats (vtc_rttm and vcm_rttm)

ChildProject/pipelines/eafbuilder.py
```diff
@@ -19,15 +21,24 @@ def create_eaf(
...
     import pympi
 
     eaf = pympi.Elan.Eaf(etf_path)
 
     ling_type = "transcription"
+    """
     eaf.add_tier("code_" + eaf_type, ling=ling_type)
     eaf.add_tier("context_" + eaf_type, ling=ling_type)
     eaf.add_tier("code_num_" + eaf_type, ling=ling_type)
+    """
```
alecristia, line 37: this is a bit where my very poor read-only skills in python are tricking me... what is the """? and why are lines 34-37 so similar to 39-40, that look like a more general version of 34-36?

alecristia: I also wonder about the ling_type parameter...

alecristia (approved):

ChildProject/pipelines/eafbuilder.py
LoannPeurey: """ surrounded blocks can be considered as comments.
Here, we thought it was better to have those three eaf tiers being children of a supertier 'SAMPLER' . This was used for the lyon-2022 annotation campaign for example. Not sure how definitive we want this behaviour to be.

ChildProject/pipelines/eafbuilder.py
LoannPeurey: This has been changed, it directly takes the LENA label as the tier name

ChildProject/pipelines/eafbuilder.py
LoannPeurey: At the moment, this has not been a problem, to keep in mind for future use

ChildProject/pipelines/eafbuilder.py
```diff
@@ -5,9 +5,11 @@
 import shutil
 
 from ChildProject.projects import ChildProject
+from ChildProject.annotations import AnnotationManager
 from ChildProject.pipelines.pipeline import Pipeline
 from ChildProject.tables import assert_dataframe, assert_columns_presence
 
+FORMAT_SPEECH = ["vtc_rttm","vcm_rttm"] #formats for which nan must be replaced with SPEECH
```
William-N-Havard, line 12: It would be cleaner to store standard formats using enum classes at some point

ChildProject/pipelines/eafbuilder.py
```diff
@@ -19,15 +21,24 @@ def create_eaf(
...
     eaf.add_tier("context_" + eaf_type, ling=ling_type)
     eaf.add_tier("code_num_" + eaf_type, ling=ling_type)
+    """
+    eaf.add_tier("SAMPLER", ling=ling_type)
+    eaf.add_tier("code_" + eaf_type, ling=ling_type, parent="SAMPLER")
+    eaf.add_tier("context_" + eaf_type, ling=ling_type, parent="SAMPLER")
+    eaf.add_tier("code_num_" + eaf_type, ling=ling_type, parent="SAMPLER")
 
     for i, ts in enumerate(timestamps_list):
         print("Creating eaf code segment # ", i + 1)
```
William-N-Havard, line 44: better to use logging.info than prints

William-N-Havard (approved):

PR #337, opened 2022-01-17
Adding a msc_type tier

marianne-m:
- [ ] tests
- [x] documentation
- [x] implementation

### DIFF
--- a/ChildProject/converters.py
+++ b/ChildProject/converters.py
@@ -472,7 +472,6 @@ class EafConverter(AnnotationConverter):
                     "vcm_type": "NA",
                     "lex_type": "NA",
                     "mwu_type": "NA",
-                    "msc_type": "NA",
                     "addressee": "NA",
                     "transcription": value if value != "0" else "0.",
                     "words": "NA",
--- a/ChildProject/converters.py
+++ b/ChildProject/converters.py
@@ -472,6 +472,7 @@ class EafConverter(AnnotationConverter):
                     "vcm_type": "NA",
                     "lex_type": "NA",
                     "mwu_type": "NA",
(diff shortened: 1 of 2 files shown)


lucasgautheron (approved):

lucasgautheron:
The tests fail because https://github.com/LAAC-LSCP/ChildProject/blob/master/tests/truth/eaf.csv lacks an empty msc_type column. One way to solve the problem is to add such a column to this csv.

Then it should be alright!

ChildProject/converters.py
```diff
@@ -472,6 +472,7 @@ def convert(filename: str, filter=None) -> pd.DataFrame:
                     "vcm_type": "NA",
                     "lex_type": "NA",
                     "mwu_type": "NA",
+                    "msc_type": "NA",
```
lucasgautheron, line 475: open question: should it be NA by default (in which case the tests need some adaptation), or simply undefined?

lucasgautheron (changes requested):
Need to decide whether msc_type is defined to NA by default or undefined by default.
Tests will require a very minor fix depending on that choice.

ChildProject/converters.py
marianne-m: I'm not sure to understand the difference between the two.
Undefined means that the column will not appear if the "msc" tier isn't in the eaf file ?

lucasgautheron: Yes, exactly!

marianne-m: In that case I feel that the undefined way is better - the "msc_type" column doesn't exist in the already imported annotations

lucasgautheron (approved):

PR #350, opened 2022-02-08
Annotations/import any eaf tier

marianne-m:
- [x] tests
- [x] documentation
- [x] implementation


lucasgautheron:
I would suggest having a `params: dict` argument for `_import_annotation` rather than `new_tiers` so we can support more parameters of that kind in the long run.

Then we can do:

`df = converter.convert(path, filter, **params)`

The signature of `Converter.convert()` should include `**kwargs`.

What do you think?

marianne-m:
Yes that's a good idea, and also ```params: dict``` in ```import_annoation```, or should I keep an explicit argument like ```new_tiers```?

Also, is there a reason to use `dict` and not `list` (and then we call `df.converter(path, filter, *params)` with `args` in the signature of `Converter.convert()`) ?

lucasgautheron (approved):

William-N-Havard (approved):

PR #362, opened 2022-02-24
check if value is in controlled vocabulary for newtiers

marianne-m:


William-N-Havard (approved):

lucasgautheron:
Few comments on this PR!

 - I was not convinced that the output belongs into `metadata` rather than `annotations/<set>`, but I think it does not matter so much!
 - I think the metadata file should contain information about the annotation set (`annotation['set']`) and the input annotation (`annotation['raw_filename']`). Maybe also the date of importation? (just like for the annotations' index).

What do you think?

lucasgautheron (changes requested):
My suggestion: add the annotation set (since different sets may have different schemes for the eaf right) and the raw filename of the eaf (for debugging purposes mostly) to the controlled voc metadata.

(All this is up to you of course! Just my advice)

William-N-Havard:
I am in favour of putting the controlled vocabulary metadata inside the metadata folder. I find it simpler to have everything stored in one place rather than in several places.

I agree it's important to save the name of the annotation set in the metadata file. But I'm not convinced we should also save the raw_filename of all the files using this annotation scheme as all the files of a given set should be using it anyway...

marianne-m:
For now the controlled vocabulary looks like this :

controlled_voc

It's a good idea to add the annotation set !
I can add a column `annotation_sets`, with a list of the sets in which the tier is used.

I'm not sure about the `raw_filename` (and `importation_date`), the table would be difficult to read...

lucasgautheron:
I agree that *ideally* all the files should share the same vocabulary within a set, but for debugging purposes it is still useful (I think) to keep track of where the information was retrieved. That does not mean keeping a duplicate entry for each file (so that should not pose any readability problem). Also this should allow to spot mismatches between annotations within a set if they occur. If you can't see a situation where this may be useful then forget about it! It's just an idea:)

PR #370, opened 2022-04-13
Patch metrics

LoannPeurey:
Modifications to the metric cli 
-add the option to add columns to include in the resulting metrics from recordings.csv and children.csv (eg child_dob, start_time).
Lines with conflicts (eg start_time when metrics are extracted for a child_id and there are multiple recordings for that child) will be outputted as NA.

-minor bugfix and updates to deprecated functions

### DIFF
--- a/ChildProject/pipelines/metrics.py
+++ b/ChildProject/pipelines/metrics.py
@@ -20,26 +20,62 @@ class Metrics(ABC):
         recordings: Union[str, List[str], pd.DataFrame] = None,
         from_time: str = None,
         to_time: str = None,
+        rec_cols: str = None,
+        child_cols: str = None,
     ):
 
         self.project = project
         self.am = ChildProject.annotations.AnnotationManager(self.project)
 
-        recording_columns = {
+        #necessary columns to construct the metrics
+        join_columns = {
             "recording_filename",
             "child_id",
             "duration",
             "session_id",
             "session_offset",
         }
-        recording_columns &= set(self.project.recordings.columns)
-
+        #get existing columns of the dataset for recordings
+        correct_cols = set(self.project.recordings.columns)
+        if rec_cols:
+            #when wanted columns specified, build the list and verify they exist (warn otherwise)
+            rec_cols=set(rec_cols.split(","))
+            for i in rec_cols:
+                if i not in correct_cols:
+                    print("Warning, specified column <{}> does not exist in recordings.csv, ignoring this column".format(i))
+            rec_cols &= correct_cols
+            #add wanted columns to the one we already get
+            join_columns.update(rec_cols)
+        self.rec_cols = rec_cols
+        
+        join_columns &= correct_cols
+        
+        #join dataset annotation with their info in recordings.csv
         self.am.annotations = self.am.annotations.merge(
-            self.project.recordings[recording_columns],
+            self.project.recordings[list(join_columns)],
             left_on="recording_filename",
             right_on="recording_filename",
         )
+        
+        #get existing columns of the dataset for children
+        correct_cols = set(self.project.children.columns)
+        if child_cols:
+            #when wanted columns specified, build the list and verify they exist (warn otherwise)
+            child_cols = set(child_cols.split(","))
+            child_cols.add("child_id")
+            for i in child_cols:
+                if i not in correct_cols:
+                    print("Warning, specified column <{}> does not exist in children.csv, ignoring this column".format(i))
+            child_cols &= correct_cols
+        self.child_cols = child_cols
 
+        #join dataset annotation with their info in children.csv
+        self.am.annotations = self.am.annotations.merge(
+            self.project.children[list(child_cols)],
+            left_on="child_id",
+            right_on="child_id",
+        )
+        
         self.by = by
         self.segments = pd.DataFrame()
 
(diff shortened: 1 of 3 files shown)


ChildProject/pipelines/metrics.py
```diff
@@ -20,26 +20,65 @@ def __init__(
...
         to_time: str = None,
+        rec_cols: str = None,
+        child_cols: str = None,
     ):
 
         self.project = project
         self.am = ChildProject.annotations.AnnotationManager(self.project)
 
-        recording_columns = {
+        #necessary columns to construct the metrics
```
alecristia, line 30: great that we start to have more in-code comments!

ChildProject/pipelines/metrics.py
```diff
@@ -20,26 +20,65 @@ def __init__(
...
             "child_id",
             "duration",
             "session_id",
             "session_offset",
         }
-        recording_columns &= set(self.project.recordings.columns)
-
+        #get existing columns of the dataset for recordings
+        correct_cols = set(self.project.recordings.columns)
+        if by not in correct_cols: exit("<{}> is not specified in this dataset, cannot extract from it".format(by))
```
alecristia, line 40: can you please add some text to the user, so that they have an idea of what they can try to do? If I read this correctly, we should say "required column not specified, please check the spelling" or something like that?

ChildProject/pipelines/metrics.py
```diff
@@ -20,26 +20,65 @@ def __init__(
...
             left_on="recording_filename",
             right_on="recording_filename",
         )
-
+        
+        #get existing columns of the dataset for children
+        correct_cols = set(self.project.children.columns)
+        if child_cols:
+            #when wanted columns specified, build the list and verify they exist (warn otherwise)
+            child_cols = set(child_cols.split(","))
```
alecristia, line 65: pls edit comment above to say "when user requests children columns, build..."

ChildProject/pipelines/metrics.py
```diff
@@ -20,26 +20,65 @@ def __init__(
...
+            child_cols.add("child_id")
+            for i in child_cols:
+                if i not in correct_cols:
+                    print("Warning, specified column <{}> does not exist in children.csv, ignoring this column".format(i))
```
alecristia, line 69: same comment as above RE warning

ChildProject/pipelines/metrics.py
```diff
@@ -20,26 +20,65 @@ def __init__(
...
+    
+            #join dataset annotation with their info in children.csv
+            self.am.annotations = self.am.annotations.merge(
+                self.project.children[list(child_cols)],
+                left_on="child_id",
+                right_on="child_id",
+            )
+        else:
+            self.child_cols = None
+
```
alecristia, line 81: if we were super nice, we'd add a check to see if the number of rows has changed before vs after the merges, which would suggest that the metrics need to be interpreted with caution. For example, imagine someone who got metrics at the child or session level, and then merges with recording columns -- they could end up with lines they believe pertain to the recordings but in fact have been collapsed across all recs (of one session) per child.

ChildProject/pipelines/metrics.py
```diff
@@ -212,10 +257,32 @@ def _process_unit(self, unit: str):
...
+                #for every unit drop the duplicates for that column
+                value=self.project.recordings[
+                        self.project.recordings[self.by] == unit
+                ][label].drop_duplicates()
+                #check that there is only one row remaining (ie this column has a unique value for that unit)
+                if len(value) == 1:
+                    metrics[label]=value.iloc[0]
+                #otherwise, leave the column as NA
+                else:
+                    metrics[label]="NA"
```
alecristia, line 285: oh i see! so you were already thinking about a situation like that I raised in my previous comment... I think this solution is more elegant than the one I proposed, which users may find frustrating

alecristia (approved):
I think most of my comments are irrelevant & this can be merged!

ChildProject/pipelines/metrics.py
LoannPeurey: lets change it to : <{}> is not specified in this dataset, cannot extract by it, change your --by option

ChildProject/pipelines/metrics.py
```diff
@@ -20,26 +20,65 @@ def __init__(
...
-
+        #get existing columns of the dataset for recordings
+        correct_cols = set(self.project.recordings.columns)
+        if by not in correct_cols: exit("<{}> is not specified in this dataset, cannot extract from it".format(by))
+        if rec_cols:
+            #when wanted columns specified, build the list and verify they exist (warn otherwise)
+            rec_cols=set(rec_cols.split(","))
+            for i in rec_cols:
+                if i not in correct_cols:
+                    print("Warning, specified column <{}> does not exist in recordings.csv, ignoring this column".format(i))
```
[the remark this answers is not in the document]
LoannPeurey: change the warning to : `Warning, requested column <{}> does not exist in recordings.csv, ignoring this column. existing columns are : {}".format(i,correct_cols)`

ChildProject/pipelines/metrics.py
LoannPeurey: `print("Warning, requested column <{}> does not exist in children.csv, ignoring this column. existing columns are : {}".format(i,correct_cols))`

ChildProject/pipelines/metrics.py
LoannPeurey: This is addressed by giving the column the NA value. Perhaps we should add a warning when that happens?

ChildProject/pipelines/metrics.py
LoannPeurey: Perhaps issue a warning along with it?

PR #376, opened 2022-06-13
Rework metrics

LoannPeurey:
The metrics pipeline needs more flexibility, the goal is to have the possibility to include whatever metric we want in the resulting csv file (from a list of available metrics or from a custom made function taking annotations and duration and outputting a value to include in the result) Old pipelines should be kept, but will work by simply giving the list of what was done in the past to the new system.
The period pipeline is going to disappear as `--period` will become an option available for every extraction and not restricted to that specific subcommand.
In order to give all the metrics wanted as parameters, a file will be used (yml ? / csv ?, maybe both). At the end of the extraction, all the parameters used will be saved in a file.
This way of handling things will come with a performance hit (as every metric is computed separately, the previous optimizations taking advantage of dependencies or computing in batch the values will not be possible anymore) and this needs to be evaluated

- [x] core of the Metrics class
- [x] list of available metrics with their function (metricsFunctions.py)
- [x] LenaMetrics and AclewMetrics backward compatibility
- [x] Usage of custom functions passed to the Metrics class to allow for custom metrics
- [x] Define the format and content of parameters file
- [x] outputs a file containing the parameters used
- [ ] ~~performance comparison~~ moved as an issue for later
- [x] Documentation
- [x] Tests
- [ ] ~~Turns count implementation~~ later implementation


LoannPeurey:
- [ ] Manage conflicts between cli arguments and parameters in the yml file (probably raise error if conflict)

LoannPeurey:
- [ ] check that recordings option works

ChildProject/pipelines/metrics.py
```diff
@@ -1,31 +1,71 @@
...
         self.project = project
         self.am = ChildProject.annotations.AnnotationManager(self.project)
+        self.threads = int(threads)
+        
+        #check that the callable column is either a callable function or a string that can be evaluated as being part of the list of metrics in ChildProject/pipelines/metricsFunctions.py
+        def check_callable(row):
+            if callable(row["callable"]): return row["callable"]
+            if isinstance(row["callable"],str):
+                try:
+                    f = eval("metfunc."+row["callable"])
```
William-N-Havard, line 45: Maybe safer do this instead of `eval':

```
f = getattr(metfunc, row['callable'])
```

ChildProject/pipelines/metrics.py
```diff
@@ -80,46 +120,262 @@ def __init__(
...
+        for i, line in self.metrics_list.iterrows():
+            curr_set = line["set"]
+            if prev_set != curr_set:
+                index, annotations = self.retrieve_segments([curr_set],row[1])
+                if index.shape[0]:
+                    duration_set = (
+                            index["range_offset"] - index["range_onset"]
+                        ).sum()
+                else : duration_set = 0
+                row[1]["duration_{}".format(line["set"])] = duration_set
```
William-N-Havard, line 180: Could you add a comment to explain what row[1] is? (and what is row[0] ?)

ChildProject/pipelines/metrics.py
```diff
@@ -152,188 +406,51 @@ def __init__(
...
+             ["avg_voc_dur_speaker",self.set,'OCH'],
+             ["avg_voc_dur_speaker",self.set,'CHI'],
+             ["wc_speaker_ph",self.set,'FEM'],
+             ["wc_speaker_ph",self.set,'MAL'],
+             ["wc_adu_ph",self.set,pd.NA],
+             ["lp_n",self.set,pd.NA],
+             ["lp_dur",self.set,pd.NA],
+             ]), columns=["callable","set","speaker"])
+
+        super().__init__(project, METRICS, by=by, recordings=recordings,period=period, from_time=from_time, to_time=to_time, rec_cols=rec_cols, child_cols=child_cols, threads=threads)
```
William-N-Havard, line 440: Break line into two or three lines

ChildProject/pipelines/metrics.py
```diff
@@ -912,6 +674,90 @@ def setup_parser(parser):
...
+            try:
+                parameters = yaml.safe_load(stream)
+                if 'parameters' in parameters: parameters = parameters['parameters']
+            except yaml.YAMLError as exc:
+                raise yaml.YAMLError("parsing of the parameters file {} failed. See above exception for more details".format(parameters_input)) from exc
+                
+        if parameters:
+            if "path" not in parameters : raise ValueError("the parameter file {} must contain at least the 'path' key specifying the path to the dataset".format(parameters_input))
+            if "destination" not in parameters : raise ValueError("the parameter file {} must contain the 'destination' key specifying the file to output the metrics to".format(parameters_input))
+            if "metrics_list" not in parameters : raise ValueError("the parameter file {} must contain the 'metrics_list' key containing the list of the desired metrics".format(parameters_input))
```
William-N-Havard, line 704: Break lines

ChildProject/projects.py
```diff
@@ -612,6 +612,13 @@ def get_recordings_from_list(
             _recordings = _recordings[
                 _recordings["recording_filename"].isin(recordings)
             ]
+            
+            if _recordings.shape[0] < len(recordings):
+                recs = pd.Series(recordings)
+                missing_recs = recs[~recs.isin(self.recordings['recording_filename'])].tolist()
+                #self.recordings[~self.recordings['recording_filename'].isin(recordings)]['recording_filename'].tolist()
+                raise ValueError("recordings {} were not found in the dataset index. Check the names and make sure they exist in 'metadata/recordings.csv'".format(missing_recs))
```
William-N-Havard, line 620: Break line

ChildProject/utils.py
```diff
@@ -45,6 +46,46 @@ def intersect_ranges(xs, ys):
...
+    d = ti1.start <= ti2.start
+    e = ti1.start <= ti2.stop
+    f = ti2.start <= ti1.stop
+    r = []
+    if c and (d and (not e or f) or not e and f) or d and not e and f : r = [TimeInterval(ti2.start,ti1.stop)]
+    elif not c and (d and (b or not a) or not a and b) or not a and b and d : r = [ti2]
+    elif not c and (not d and (not e and not f or e) or e and not f) or not d and e and not f : r = [TimeInterval(ti1.start,ti2.stop)]
+    elif c and (not d and (not a and not b or a) or a and not b) or a and not b and not d : r = [ti1]
+    # !! here the expression was simplified because previous statements should already have caught their relevant cases (ie this statement should always be last or changed)
+    elif not a and (not b or e) or d and e and f : r = [TimeInterval(ti1.start,ti2.stop),TimeInterval(ti2.start,ti1.stop)]
```
William-N-Havard, line 81: Perhaps it should be commented on a little more ;)

William-N-Havard (approved):

ChildProject/utils.py
LoannPeurey: yes, probably... Added some line to at least have some info on what it does but I don't think explaining the whole process that lead to those boolean expression is really useful. Or we could change the whole function to work in a more comprehensible (and probably less optimized) way ?

ChildProject/pipelines/metrics.py
LoannPeurey: for sure, knew there was a better way of doing this!
 7daaba6

ChildProject/pipelines/metrics.py
LoannPeurey: 7daaba6

ChildProject/utils.py
William-N-Havard: I found a solution that is a little bit more verbose but clearer in my opinion (I haven't tested it thoroughly though). 
Here is the dump of my Jupyter notebook:

```python
from datetime import datetime
```


```python
class TimeInterval:
    def __init__(self, start : datetime, stop : datetime):
        #remove the day/month/year component
        self.start = start.replace(year=1900, month=1, day=1)
        self.stop = stop.replace(year=1900, month=1, day=1)

    def length(self):
        return self.stop - self.start

    def __repr__(self):
        return "TimeInterval([{}, {}])".format(self.start, self.stop)
```


```python
# Utility functions
def overlap(onset, offset, target_onset, target_offset):
    return max(0, min(offset, target_offset) - max(onset, target_onset))

def overlaps(onset, offset, target_onset, target_offset):
    return bool(overlap(onset, offset, target_onset, target_offset))

def time_as_decimal(datetime_):
    return datetime_.hour + datetime_.minute / 60 + datetime_.second / 3600

def break_time(ti : TimeInterval):
    ti_ = []
    if ti.start < ti.stop:
        ti_.append((ti.start, ti.stop))
    else:
        # Break at midnight
        ti_.extend([(ti.start, datetime(1900,1,1,23,59)), 
                    (datetime(1900,1,1,0,0), ti.stop)])
    return ti_
```


```python
# Should break
break_time(TimeInterval(datetime(1900,1,1,21,4),datetime(1900,1,1,10,36)))
```




    [(datetime.datetime(1900, 1, 1, 21, 4), datetime.datetime(1900, 1, 1, 23, 59)),
     (datetime.datetime(1900, 1, 1, 0, 0), datetime.datetime(1900, 1, 1, 10, 36))]




```python
# Shouldn't break
break_time(TimeInterval(datetime(1900,1,1,8,57),datetime(1900,1,1,22,1)))
```




    [(datetime.datetime(1900, 1, 1, 8, 57), datetime.datetime(1900, 1, 1, 22, 1))]




```python
# Convert to decimal so as to be usable by overlap (maybe change this later)
time_as_decimal(datetime(1900,1,1,22,59,59))
```




    22.999722222222225




```python
def time_intervals_intersect(ti1 : TimeInterval, ti2 : TimeInterval):
    ti1_broken = break_time(ti1)
    ti2_broken = break_time(ti2)
    
    olaps = []
    for start1, stop1 in ti1_broken:
        st1 = time_as_decimal(start1)
        sp1 = time_as_decimal(stop1)
        
        for start2, stop2 in ti2_broken:
            st2=time_as_decimal(start2)
            sp2=time_as_decimal(stop2)

            if overlap(st1,sp1,st2,sp2):
                olaps.append((max(start1,start2), min(stop1,stop2)))
    return olaps
```


```python
time_intervals_intersect(TimeInterval(datetime(1900,1,1,8,57),datetime(1900,1,1,21,4)),TimeInterval(datetime(1900,1,1,10,36),datetime(1900,1,1,22,1)))
```




    [(datetime.datetime(1900, 1, 1, 10, 36), datetime.datetime(1900, 1, 1, 21, 4))]




```python
time_intervals_intersect(TimeInterval(datetime(1900,1,1,8,57),datetime(1900,1,1,22,1)),TimeInterval(datetime(1900,1,1,21,4),datetime(1900,1,1,10,36)))
```




    [(datetime.datetime(1900, 1, 1, 21, 4), datetime.datetime(1900, 1, 1, 22, 1)),
     (datetime.datetime(1900, 1, 1, 8, 57), datetime.datetime(1900, 1, 1, 10, 36))]

William-N-Havard: What you could do too is use `time` instead of `datetime` with a date set to 1900  (which is a bit hacky in my opinion)

```
        self.start = time(start.hour, start.minute, start.second)
        self.stop = time(stop.hour, stop.minute, stop.second)
```

ChildProject/pipelines/metrics.py
```diff
@@ -80,46 +120,266 @@ def __init__(
...
+                annotations, TimeInterval(self.from_time,self.to_time))
+        elif self.period:
+            st_hour = row["period_start"]
+            end_hour = row["period_end"]
+            matches = self.am.get_within_time_range(
+                    annotations, TimeInterval(st_hour,end_hour))
+        else:
+            matches = annotations
 
         try:
```
kalenkovich, line 247: What are `segments`? Do exceptions here correspond to "no segments for these annotations"?

ChildProject/pipelines/metrics.py
```diff
@@ -1,31 +1,71 @@
...
 from ChildProject.pipelines.pipeline import Pipeline
 
-pipelines = {}
+import ChildProject.pipelines.metricsFunctions as metfunc #used by eval
+from ..utils import TimeInterval, time_intervals_intersect
 
+pipelines = {}
 
 class Metrics(ABC):
     def __init__(
```
kalenkovich, line 22: I think a docstring might be useful here. I think I understood what all arguments do but I am not sure.

ChildProject/pipelines/metrics.py
```diff
@@ -152,188 +410,53 @@ def __init__(
         self,
         project: ChildProject.projects.ChildProject,
         set: str,
```
kalenkovich, line 412: Why is `set` an argument for a specific type of metrics, like LENA here? Could there be multiple sets of LENA annotations? Or can the user choose a custom name for the set? Sorry, I am still a bit confused about the data structure and sets.

BTW, is there a glossary of sorts? With things like "recording", "set", "pipeline", etc.? Obviously, you guys all know that stuff but it might be helpful to newcomers.

ChildProject/pipelines/metrics.py
LoannPeurey: annotations represent a period of the audio file that is annotated (e.g. a row of annotations having set = 'its' , range_onset = '0' , range_offset = '3600,000' is the first hour of the recording annotated in the its set)
Segments are the actual vocalizations, so if you do `get_segments` on the example of annotation above, it will return a DataFrame containing all the vocalizations (segment_onset/offset, speaker_type [etc](https://childproject.readthedocs.io/en/latest/format.html#id10) ) that were detected by lena it the first hour of the audio.
And yes, the exception is for when no vocalizations exist for this anotated period.

ChildProject/pipelines/metrics.py
LoannPeurey: The use of 'its' as the name of the lena set is what we do by default but yes, this allows people to give the name they want to their lena set.
At the moment, we don't really have a glossary, that sounds like a really good idea to improve communication 👍 . We often ourselves get mixed up between words like "segments"/ "annotations" ...

ChildProject/pipelines/metrics.py
LoannPeurey: 👍 will add it !

If you have any doubt, the class that inherit from this class (e.g. LenaMetrics) have usually all of those arguments (and some more) . So their behavious is the same

ChildProject/pipelines/metrics.py
kalenkovich: Got it, thanks!

Could catching all exceptions lead to problems here? I am not assuming this is necessarily a problem but maybe worth a comment on why it is a good idea in this case?

ChildProject/projects.py
```diff
@@ -612,6 +612,15 @@ def get_recordings_from_list(
...
             ]
+            
+            if _recordings.shape[0] < len(recordings):
+                recs = pd.Series(recordings)
+                missing_recs = recs[~recs.isin(self.recordings['recording_filename'])].tolist()
+                #self.recordings[~self.recordings['recording_filename'].isin(recordings)]['recording_filename'].tolist()
+                raise ValueError("recordings {} were not found in the dataset index.\
+                                 Check the names and make sure they exist in\
+                                 'metadata/recordings.csv'".format(missing_recs))
+
```
kalenkovich, line 623: That is unrelated to the new metrics pipeline, right?

docs/source/metrics.rst
```diff
@@ -6,48 +6,36 @@ Overview
...
-    cp_n,canonical proportion = canonical /(can+noncan) based on number of vocalizations,ACLEW
-    lp_dur,linguistic proportion = (speech)/(cry+speech) based on duration of vocalizations,"ACLEW,LENA"
-    cp_dur,canonical proportion = canonical /(can+noncan) based on duration of vocalizations,ACLEW
-
-.. note::
-
-    Average rates are expressed in counts/hour (for events) or in seconds/hour (for durations).
+.. warning::
+
+    Be aware that numerous metrics are rates (every metric ending with 'ph' is) and not absolute counts!
```
kalenkovich, line 32: Thanks for that comment! :-)

kalenkovich (commented):
New code and tests make sense to me. @LoannPeurey, would you like me to check anything specific in more detail?

ChildProject/pipelines/metrics.py
```diff
@@ -80,46 +120,266 @@ def __init__(
...
         try:
-            segments = self.am.get_segments(annotations)
+            segments = self.am.get_segments(matches)
         except Exception as e:
-            print(str(e))
+            #print(str(e))
+            #no annotations for that unit
             return pd.DataFrame(), pd.DataFrame()
 
         # prevent overflows
```
kalenkovich, line 254: (Obviously, has nothing with the new pipeline, so feel free to ignore it completely) Why is there a risk of overflow? Is that because the total duration might be too big an int? Wouldn't it be less than the maximum offset though?

docs/source/metrics.rst
LoannPeurey: ;-)

ChildProject/projects.py
LoannPeurey: Yes, it is related to the `--reordings` option that appears in numerous pipelines. If the recordings given did not exist in the dataset, the command would just continue and not print anything, so raising an error semmed like a better way of handling things

ChildProject/pipelines/metrics.py
LoannPeurey: To be honest, I don't know any better. I checked the commits and that comment was already present in the first version of the file, with almost the same code. As for reaching a too high int value, I can't see that happening as I think that corresponds to weeks of continuous recordings...

LoannPeurey:
> New code and tests make sense to me. @LoannPeurey, would you like me to check anything specific in more detail?

Next step being the addition of the lena CTC CVC to our supported metrics list, is that clear for you how we do it?

ChildProject/pipelines/metrics.py
LoannPeurey: I think you are right, it is probably not justified to catch exceptions for that particular case, the function should be able to have an empty dataframe output there.
(e57fdb2bd)

kalenkovich:
> > New code and tests make sense to me. @LoannPeurey, would you like me to check anything specific in more detail?
> 
> Next step being the addition of the lena CTC CVC to our supported metrics list, is that clear for you how we do it?

Something like this?

```
@metricFunction(args={}, cols={"lena_ctc"})
def lena_ctc(annotations: pd.DataFrame, max_distance_ms: int = 1, **kwargs):
   <pretty much a copy of the current ctc code>
```

Then "lena_ctc" should be supplied in the input dataframe.

LoannPeurey:
The main difference is that we don't have a "lena_ctc" column extracted from its files.
We store the columns:
- lena_speaker
- lena_block_number
- lena_block_type
- lena_conv_status
- lena_response_count
- lena_conv_turn_type
- lena_conv_floor_type
- utterances_count
- utterances_length

So CTC should be computed from some of this columns I think.
You can see what converted its files are from [here](https://gin.g-node.org/EL1000/bergelson/src/main/annotations/its/converted/123417-0008_1_0_45871800.csv) for example. And the description of the column content [here](https://childproject.readthedocs.io/en/latest/format.html#id8)

I see that you have added the `max_distance_ms` argument, foes lena allow to change that? If we can include it, it should be as a kwarg, the duration arg however should alway be here, even if not used because it is passed to all metrics.

So closer to: 
```python
@metricFunction(args={}, cols={"lena_conv_status","lena_conv_turn_type" }) #put relevant columns
def lena_ctc(annotations: pd.DataFrame, duration: int, **kwargs):
   max_distance_ms = kwargs["max_distance_ms"]
   <pretty much a copy of the current ctc code>
```

kalenkovich:
Hi, @LoannPeurey! I totally misunderstood your question, to be honest. I thought you were checking if I understood the new metrics logic 🤦 And obviously I misread the purpose of the `columns` argument of the decorator function 🤦 🤦 Also I forgot that we were trying to get ctc as calculated by LENA, not calculate ctc from its files the same way it is calculated from the VTC files. Hence, the `max_distance_ms: int = 1` argument. 🤦🤦🤦

As for the actual question, I would extract one more column from its by adding the following to `converters.py`

```python
class ItsConverter(AnnotationConverter):
    ...
                lena_cumulative_ctc = conversation_info[2]
```

That field is empty except for the segments which are the last part of a given conversational turn, therefore, it has to be forward-filled and then zero-filled in the beginning. I would probably do it at the conversion stage with something like

```python
        df = pd.DataFrame(segments)
        
        # lena_cumulative_ctc is NA for any segment where ctc hasn't changed, let's fill the holes
        df['lena_cumulative_ctc'] = df.lena_cumulative_ctc.ffill().fillna(0)

        return df
```

After that, for any interval we would take `max(lena_cumulative_ctc) - min(lena_cumulative_ctc)`:

```python
@metricFunction(columns={"lena_cumulative_ctc"}, emptyValue = 0)
def lena_ctc(annotations: pd.DataFrame, duration: int, **kwargs):
    """Conversational turn count (ctc) as calculated by LENA."""
    return annotations.lena_cumulative_ctc.max() - annotations.lena_cumulative_ctc.min()
```

ChildProject/pipelines/metricsFunctions.py
```diff
@@ -0,0 +1,301 @@
...
+        def new_func(annotations: pd.DataFrame, duration: int, **kwargs):
+            for arg in args:
+                if arg not in kwargs : raise ValueError('{} metric needs an argument <{}>'.format(function.__name__,arg))
+            metname = name
+            if not name : metname = function.__name__            
+            for arg in kwargs:
+                metname = re.sub(arg , str(kwargs[arg]).lower(),metname)
+            if annotations.shape[0]:
+                for column in columns:
+                    if column not in annotations.columns : raise ValueError(MISSING_COLUMNS.format(annotations['set'].iloc[0],column,'voc_speaker_ph'))
```
kalenkovich, line 44: I think `.format(annotations['set'].iloc[0],column,'voc_speaker_ph'))` should be `.format(annotations['set'].iloc[0], column, metname))

kalenkovich: Oh, and it is obviously a matter of personal preference but at first I couldn't figure out what `metname` is and what is so meta about it :smile: I would personally go with something like `metric_name`.

kalenkovich, line 44: Totally nit-picky but as a user I'd prefer to learn about all the missing columns at once, so something like:

```python
missing_columns = columns - set(annotations.columns)
if missing_columns:
    raise ValueError(MISSING_COLUMNS.format(annotations['set'].iloc[0], missing_columns, metname))
```

LoannPeurey: You are right, this mistake slipped through the net, We corrected it shortly after the merge :+1:  
I'll keep in mind the variable name change and will do it shortly.

LoannPeurey: :+1:

LoannPeurey:
Since we already merged this branch and went on to new additions, I suggest you check out the #394 PR, @kalenkovich I'll add you as a reviewer there, so check the changes if you want (and your review will probably be useful). Just know that this will be merged and hopefully released at the beginning of next week regardless.

ChildProject/pipelines/metricsFunctions.py
LoannPeurey: f61066db

LoannPeurey: f61066db

LoannPeurey:
> Hi, @LoannPeurey! I totally misunderstood your question, to be honest. I thought you were checking if I understood the new metrics logic 🤦 And obviously I misread the purpose of the `columns` argument of the decorator function 🤦 🤦 Also I forgot that we were trying to get ctc as calculated by LENA, not calculate ctc from its files the same way it is calculated from the VTC files. Hence, the `max_distance_ms: int = 1` argument. 🤦🤦🤦
> 
> As for the actual question, I would extract one more column from its by adding the following to `converters.py`
> 
> ```python
> class ItsConverter(AnnotationConverter):
>     ...
>                 lena_cumulative_ctc = conversation_info[2]
> ```
> 
> That field is empty except for the segments which are the last part of a given conversational turn, therefore, it has to be forward-filled and then zero-filled in the beginning. I would probably do it at the conversion stage with something like
> 
> ```python
>         df = pd.DataFrame(segments)
>         
>         # lena_cumulative_ctc is NA for any segment where ctc hasn't changed, let's fill the holes
>         df['lena_cumulative_ctc'] = df.lena_cumulative_ctc.ffill().fillna(0)
> 
>         return df
> ```
> 
> After that, for any interval we would take `max(lena_cumulative_ctc) - min(lena_cumulative_ctc)`:
> 
> ```python
> @metricFunction(columns={"lena_cumulative_ctc"}, emptyValue = 0)
> def lena_ctc(annotations: pd.DataFrame, duration: int, **kwargs):
>     """Conversational turn count (ctc) as calculated by LENA."""
>     return annotations.lena_cumulative_ctc.max() - annotations.lena_cumulative_ctc.min()
> ```

Understood, if we have to import a new column from the its, it means that this will break backward compatibility and we will need to reimport most of our datasets

kalenkovich:
> > Hi, @LoannPeurey! I totally misunderstood your question, to be honest. I thought you were checking if I understood the new metrics logic 🤦 And obviously I misread the purpose of the `columns` argument of the decorator function 🤦 🤦 Also I forgot that we were trying to get ctc as calculated by LENA, not calculate ctc from its files the same way it is calculated from the VTC files. Hence, the `max_distance_ms: int = 1` argument. 🤦🤦🤦
> > As for the actual question, I would extract one more column from its by adding the following to `converters.py`
> > ```python
> > class ItsConverter(AnnotationConverter):
> >     ...
> >                 lena_cumulative_ctc = conversation_info[2]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > That field is empty except for the segments which are the last part of a given conversational turn, therefore, it has to be forward-filled and then zero-filled in the beginning. I would probably do it at the conversion stage with something like
> > ```python
> >         df = pd.DataFrame(segments)
> >         
> >         # lena_cumulative_ctc is NA for any segment where ctc hasn't changed, let's fill the holes
> >         df['lena_cumulative_ctc'] = df.lena_cumulative_ctc.ffill().fillna(0)
> > 
> >         return df
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > After that, for any interval we would take `max(lena_cumulative_ctc) - min(lena_cumulative_ctc)`:
> > ```python
> > @metricFunction(columns={"lena_cumulative_ctc"}, emptyValue = 0)
> > def lena_ctc(annotations: pd.DataFrame, duration: int, **kwargs):
> >     """Conversational turn count (ctc) as calculated by LENA."""
> >     return annotations.lena_cumulative_ctc.max() - annotations.lena_cumulative_ctc.min()
> > ```
> 
> Understood, if we have to import a new column from the its, it means that this will break backward compatibility and we will need to reimport most of our datasets

Isn't that a common situation where you need some extra information from the raw files? Just curious, because in this case, it is probably unnecessary since your calculation should yield the same numbers as mine.

LoannPeurey:
> > > Hi, @LoannPeurey! I totally misunderstood your question, to be honest. I thought you were checking if I understood the new metrics logic 🤦 And obviously I misread the purpose of the `columns` argument of the decorator function 🤦 🤦 Also I forgot that we were trying to get ctc as calculated by LENA, not calculate ctc from its files the same way it is calculated from the VTC files. Hence, the `max_distance_ms: int = 1` argument. 🤦🤦🤦
> > > As for the actual question, I would extract one more column from its by adding the following to `converters.py`
> > > ```python
> > > class ItsConverter(AnnotationConverter):
> > >     ...
> > >                 lena_cumulative_ctc = conversation_info[2]
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > That field is empty except for the segments which are the last part of a given conversational turn, therefore, it has to be forward-filled and then zero-filled in the beginning. I would probably do it at the conversion stage with something like
> > > ```python
> > >         df = pd.DataFrame(segments)
> > >         
> > >         # lena_cumulative_ctc is NA for any segment where ctc hasn't changed, let's fill the holes
> > >         df['lena_cumulative_ctc'] = df.lena_cumulative_ctc.ffill().fillna(0)
> > > 
> > >         return df
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > After that, for any interval we would take `max(lena_cumulative_ctc) - min(lena_cumulative_ctc)`:
> > > ```python
> > > @metricFunction(columns={"lena_cumulative_ctc"}, emptyValue = 0)
> > > def lena_ctc(annotations: pd.DataFrame, duration: int, **kwargs):
> > >     """Conversational turn count (ctc) as calculated by LENA."""
> > >     return annotations.lena_cumulative_ctc.max() - annotations.lena_cumulative_ctc.min()
> > > ```
> > 
> > 
> > Understood, if we have to import a new column from the its, it means that this will break backward compatibility and we will need to reimport most of our datasets
> 
> Isn't that a common situation where you need some extra information from the raw files? Just curious, because in this case, it is probably unnecessary since your calculation should yield the same numbers as mine.

I am not sure if that has occured in the past. I am far from being an expert on lena's file so I am not sure how exhaustive our importation is. But in most other file types, I think the importation aims at getting all the info converted.

PR #394, opened 2022-08-08
Metrics/segments

LoannPeurey:
Add the --segments option to the metrics pipeline.
This option extracts metrics only on the required segments in a dataFrame having 'recording_filename', 'segment_onset' and 'segment_offset'.
Use this option with `--by segments`
- [X] implementation
- [X] tests
- [X] docstrings

### DIFF
--- a/ChildProject/pipelines/metrics.py
+++ b/ChildProject/pipelines/metrics.py
@@ -27,7 +27,7 @@ class Metrics(ABC):
     :type project: ChildProject.projects.ChildProject
     :param metrics_list: pandas DataFrame containing the desired metrics (metrics functions are in metricsFunctions.py)
     :type metrics_list: pd.DataFrame
-    :param by: unit to extract metric from (recording_filename, experiment, child_id, session_id, segments), defaults to 'recording_filename', use 'segments' to use the segments argument
+    :param by: unit to extract metric from (recording_filename, experiment, child_id, session_id, segments), defaults to 'recording_filename', 'segments' is mandatory if passing the segments argument
     :type by: str, optional
     :param recordings: recordings to sample from; if None, all recordings will be sampled, defaults to None
     :type recordings: Union[str, List[str], pd.DataFrame], optional
@@ -41,7 +41,7 @@ class Metrics(ABC):
     :type child_cols: str, optional
     :param period: time units to aggregate (optional); equivalent to ``pandas.Grouper`` freq argument.
     :type period: str, optional
-    :param segments: DataFrame or path to csv file of the segments to extract from, containing 'recording_filename', 'segment_onset' and 'segment_offset' columns. To use this option, the <by> option must be set to 'segments', can not be used along with options [recordings,period,from_time,to_time].
+    :param segments: DataFrame or path to csv file of the segments to extract from, containing 'recording_filename', 'segment_onset' and 'segment_offset' columns. To use this option, the option must be set to 'segments'. Also, this option cannot be combined with options [recordings,period,from_time,to_time].
     :type segments: Union[str, pd.DataFrame], optional
     :param threads: amount of threads to run on, defaults to 1
     :type threads: int, optional
@@ -153,7 +153,7 @@ class Metrics(ABC):
         self.recordings = Pipeline.recordings_from_list(recordings)
         
         if segments is not None:
-            assert by == 'segments' and period is None and recordings is None and from_time is None and to_time is None, "the <segments> option can not be used with options [period,recordings,from_time,to_time], and <by> should be set to 'segments'"
+            assert by == 'segments' and period is None and recordings is None and from_time is None and to_time is None, "the <segments> option can not be combined with options [period,recordings,from_time,to_time], and <by> should be set to 'segments'"
             if isinstance(segments, pd.DataFrame):
                 self.segments = segments
             else:
@@ -396,11 +396,11 @@ class CustomMetrics(Metrics):
     :type rec_cols: str, optional
     :param child_cols: comma separated columns from children.csv to include in the outputted metrics (optional), None by default
     :type child_cols: str, optional
-    :param by: units to sample from, defaults to 'recording_filename', use 'segments' to use the segments argument
+    :param by: unit to extract metric from (recording_filename, experiment, child_id, session_id, segments), defaults to 'recording_filename', 'segments' is mandatory if passing the segments argument
     :type by: str, optional
     :param period: time units to aggregate (optional); equivalent to ``pandas.Grouper`` freq argument.
     :type period: str, optional
-    :param segments: DataFrame or path to csv file of the segments to extract from, containing 'recording_filename', 'segment_onset' and 'segment_offset' columns. To use this option, the <by> option must be set to 'segments', can not be used along with options [recordings,period,from_time,to_time].
+    :param segments: DataFrame or path to csv file of the segments to extract from, containing 'recording_filename', 'segment_onset' and 'segment_offset' columns. To use this option, the option must be set to 'segments'. Also, this option cannot be combined with options [recordings,period,from_time,to_time].
     :type segments: Union[str, pd.DataFrame], optional
     :param threads: amount of threads to run on, defaults to 1
     :type threads: int, optional
@@ -454,11 +454,11 @@ class LenaMetrics(Metrics):
     :type rec_cols: str, optional
     :param child_cols: comma separated columns from children.csv to include in the outputted metrics (optional), None by default
     :type child_cols: str, optional
-    :param by: units to sample from, defaults to 'recording_filename', use 'segments' to use the segments argument
+    :param by: unit to extract metric from (recording_filename, experiment, child_id, session_id, segments), defaults to 'recording_filename', 'segments' is mandatory if passing the segments argument
     :type by: str, optional
     :param period: time units to aggregate (optional); equivalent to ``pandas.Grouper`` freq argument.
     :type period: str, optional
-    :param segments: DataFrame or path to csv file of the segments to extract from, containing 'recording_filename', 'segment_onset' and 'segment_offset' columns. To use this option, the <by> option must be set to 'segments', can not be used along with options [recordings,period,from_time,to_time].
+    :param segments: DataFrame or path to csv file of the segments to extract from, containing 'recording_filename', 'segment_onset' and 'segment_offset' columns. To use this option, the option must be set to 'segments'. Also, this option cannot be combined with options [recordings,period,from_time,to_time].
     :type segments: Union[str, pd.DataFrame], optional
     :param threads: amount of threads to run on, defaults to 1
     :type threads: int, optional
@@ -545,11 +545,11 @@ class AclewMetrics(Metrics):
     :type rec_cols: str, optional
     :param child_cols: comma separated columns from children.csv to include in the outputted metrics (optional), None by default
     :type child_cols: str, optional
-    :param by: units to sample from, defaults to 'recording_filename'
+    :param by: unit to extract metric from (recording_filename, experiment, child_id, session_id, segments), defaults to 'recording_filename', 'segments' is mandatory if passing the segments argument
     :type by: str, optional
     :param period: time units to aggregate (optional); equivalent to ``pandas.Grouper`` freq argument.
     :type period: str, optional
-    :param segments: DataFrame or path to csv file of the segments to extract from, containing 'recording_filename', 'segment_onset' and 'segment_offset' columns. To use this option, the <by> option must be set to 'segments', can not be used along with options [recordings,period,from_time,to_time].
+    :param segments: DataFrame or path to csv file of the segments to extract from, containing 'recording_filename', 'segment_onset' and 'segment_offset' columns. To use this option, the option must be set to 'segments'. Also, this option cannot be combined with options [recordings,period,from_time,to_time].
     :type segments: Union[str, pd.DataFrame], optional
     :param threads: amount of threads to run on, defaults to 1
     :type threads: int, optional
--- a/ChildProject/pipelines/metrics.py
+++ b/ChildProject/pipelines/metrics.py
@@ -13,7 +13,8 @@ from git.exc import InvalidGitRepositoryError
 import ChildProject
 from ChildProject.pipelines.pipeline import Pipeline
 
-import ChildProject.pipelines.metricsFunctions as metfunc #used by eval
+from ChildProject.tables import assert_dataframe, assert_columns_presence
+import ChildProject.pipelines.metricsFunctions as metfunc
 from ..utils import TimeInterval, time_intervals_intersect
 
 pipelines = {}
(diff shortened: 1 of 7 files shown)


ChildProject/pipelines/metrics.py
```diff
@@ -26,7 +27,7 @@ class Metrics(ABC):
     :type project: ChildProject.projects.ChildProject
     :param metrics_list: pandas DataFrame containing the desired metrics (metrics functions are in metricsFunctions.py)
     :type metrics_list: pd.DataFrame
-    :param by: unit to extract metric from (recording_filename, experiment, child_id, session_id), defaults to 'recording_filename'
+    :param by: unit to extract metric from (recording_filename, experiment, child_id, session_id, segments), defaults to 'recording_filename', use 'segments' to use the segments argument
```
alecristia, line 30: should we have a different explanation for segments here? perhaps something like: "'segments' is mandatory if passing the segments (-s) argument"

ChildProject/pipelines/metrics.py
```diff
@@ -40,6 +41,8 @@ class Metrics(ABC):
     :type child_cols: str, optional
     :param period: time units to aggregate (optional); equivalent to ``pandas.Grouper`` freq argument.
     :type period: str, optional
+    :param segments: DataFrame or path to csv file of the segments to extract from, containing 'recording_filename', 'segment_onset' and 'segment_offset' columns. To use this option, the <by> option must be set to 'segments', can not be used along with options [recordings,period,from_time,to_time].
```
alecristia, line 44: " To use this option, the <by> option must be set to 'segments'. Also, this option cannot be combined with options [recordings,period,from_time,to_time]."

ChildProject/pipelines/metrics.py
```diff
@@ -145,36 +149,50 @@ def check_callable(row):
...
-            self.periods = pd.interval_range(
-                start=datetime.datetime(1900, 1, 1, 0, 0, 0, 0),
-                end=datetime.datetime(1900, 1, 2, 0, 0, 0, 0),
-                freq=self.period,
-                closed="both",
-            )
-            self.periods= pd.DataFrame(self.periods.to_tuples().to_list(),columns=['period_start','period_end'])
-        
-        self.segments = pd.DataFrame()
-
```
alecristia, line 160: was this deletion intentional?

ChildProject/pipelines/metrics.py
```diff
@@ -145,36 +149,50 @@ def check_callable(row):
...
+        self.segments = segments   
         self.recordings = Pipeline.recordings_from_list(recordings)
         
-        if from_time:
-            try:
-                self.from_time = datetime.datetime.strptime(from_time, "%H:%M:%S")
-            except:
-                raise ValueError(f"invalid value for from_time ('{from_time}'); should have HH:MM:SS format instead")
+        if segments is not None:
+            assert by == 'segments' and period is None and recordings is None and from_time is None and to_time is None, "the <segments> option can not be used with options [period,recordings,from_time,to_time], and <by> should be set to 'segments'"
```
alecristia, line 156: perhaps instead tell user:
"the <segments> option cannot be combined with options [period,recordings,from_time,to_time], ...."

tests/test_metrics.py
```diff
@@ -46,7 +45,19 @@ def test_failures(project):
...
+        lm = pd.DataFrame(np.array(
+            [["voc_speaker","segments_vtc",'FEM'],         
+             ]), columns=["callable","set","speaker"])
+        m = Metrics(project, lm,  segments="unknown")
+    except ValueError as e:
+        exception_caught = True
+
+    assert (
+        exception_caught == True
+    ), "Metrics failed to throw an exception despite having the segments argument and by having a value different than 'recording_filename'"
```
alecristia, line 60: oh this is where checks happen, is it??

alecristia: and it even checks for recs being valid -- cool!

alecristia: or maybe not? maybe this is just for the test with the data provided below, right?

alecristia (changes requested):
just one thing: checks for user passing an appropriate segments.csv; otherwise it looks good!

ChildProject/pipelines/metrics.py
```diff
@@ -145,36 +149,50 @@ def check_callable(row):
...
+                self.segments = segments
+            else:
+                self.segments = pd.read_csv(segments)
+
+            assert_dataframe("segments", self.segments, not_empty=True)
+            assert_columns_presence(
+                "segments",
+                self.segments,
+                {"recording_filename", "segment_onset", "segment_offset"},
+            )
```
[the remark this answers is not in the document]
LoannPeurey: This line checks that the dataframe has at least those 3 required columns and fails when they are not present. However, we don't do a 'datatype' check at the moment to check that recording are strings and onset/offset integers.

ChildProject/pipelines/metricsFunctions.py
```diff
@@ -51,6 +51,14 @@ def new_func(annotations: pd.DataFrame, duration: int, **kwargs):
...
             
 
+@metricFunction({"speaker"},{"speaker_type"}) 
+def voc_speaker(annotations: pd.DataFrame, duration: int, **kwargs):
+    """number of vocalizations for a given speaker type
+    
+    Required keyword arguments:
+        - speaker : speaker_type to use
+    """
+    return annotations[annotations["speaker_type"]== kwargs["speaker"]].shape[0]
```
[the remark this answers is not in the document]
LoannPeurey: I added in the available metrics the voc_speaker metrics because I figured it made sense to have it available when extracting from segments (as when doing segment extraction, it is unlikely you want rates instead of actual count numbers

tests/test_metrics.py
LoannPeurey: This is indeed just the automated tests to make sure the command fails when the arguments are not correct.
But the check for recordings being in the dataset is actually done during the extraction here:
https://github.com/LAAC-LSCP/ChildProject/blob/e3dc054e968d4217b60a9b384c205907d2bfc3f7/ChildProject/pipelines/metrics.py#L317

ChildProject/pipelines/metrics.py
LoannPeurey: We don't include info about CLI arguments (--segments -d) because this is the docstring of the function, not the help of the CLI

ChildProject/pipelines/metrics.py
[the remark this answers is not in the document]
LoannPeurey: datatype is now checked with 4cad34d

ChildProject/pipelines/metricsFunctions.py
```diff
@@ -297,4 +308,13 @@ def cp_dur(annotations: pd.DataFrame, duration: int, **kwargs):
...
+    return value
+
+@metricFunction(set(),{"lena_conv_turn_type"}) 
+def lena_CVC(annotations: pd.DataFrame, duration: int, **kwargs):
+    """number of child vocalizations according to LENA's extraction
+    
+    Required keyword arguments:
+    """
+    conv_types = {'TIMR', 'TIFR'}
+    return annotations[annotations["lena_conv_turn_type"].isin(conv_types)].shape[0]
```
kalenkovich, line 320: I wonder if our implementations of lena_CVC (mine is described [here](https://github.com/LAAC-LSCP/ChildProject/pull/376#issuecomment-1212088055)) yield the same results. I'll check early next week. This pr will have already been merged, so I'll just open an issue if the result differ.

ChildProject/tables.py
```diff
@@ -13,7 +13,13 @@ def __init__(self, name: str, missing: Set):
         super().__init__(
             f"dataframe {name} misses the following required columns: {missing}"
         )
+        
+class IncorrectDtypeException(Exception):
+    def __init__(self, error: str):
 
+        super().__init__(
+            error
+        )
```
kalenkovich, line 22: This is a tiny bit confusing to me. Was `__init__` overridden here, so that `IncorrectDtypeException` will only accept a single argument?

ChildProject/pipelines/metrics.py
```diff
@@ -40,6 +41,8 @@ class Metrics(ABC):
     :type child_cols: str, optional
     :param period: time units to aggregate (optional); equivalent to ``pandas.Grouper`` freq argument.
     :type period: str, optional
+    :param segments: DataFrame or path to csv file of the segments to extract from, containing 'recording_filename', 'segment_onset' and 'segment_offset' columns. To use this option, the option must be set to 'segments'. Also, this option cannot be combined with options [recordings,period,from_time,to_time].
+    :type segments: Union[str, pd.DataFrame], optional
```
kalenkovich, line 45: I think I got lost in the terminology again. I thought that _segments_ referred to intervals with individual annotated utterances while _annotations_ referred to the time intervals that might contain _segments_. But here _segments_ refers to what I thought were _annotations_. Could you please explain the difference to me again?

kalenkovich (commented):
- A few places where I simply didn't understand what was happening.
- Your way to calculate lena_CTC is different from mine. I'll check the difference in the results - if any - on Monday.

Other than that, everything seems good.

ChildProject/tables.py
kalenkovich: Also, `error` is an unusual (to me) name for the error message. I've mostly seen `message` or `msg`.

ChildProject/pipelines/metrics.py
alecristia: sorry about that, in our terminology, segment = stretch of audio

So you can have an utterance that some human or machine has split up into three segments -- the word segment only means that they have a start & end time, it doesn't tell you that eg they are internally consistent (in the sense that I think you were thinking about, perhaps segment = phone, or segment = a single cry)

AND you can have a recording that has been split into parts, for instance, sections in which humans have carried out annotation

ChildProject/pipelines/metricsFunctions.py
alecristia: it would be great to check!
Here's the relevant its section:
```
                <Segment spkr="MAN" average_dB="-34.17" peak_dB="-24.66" conversationInfo="|EC|9|4|2|CIC|TIMR|FI|" maleAdultWordCnt="8.99" maleAdultNonSpeechLen="P0.00S" maleAdultUttCnt="0" maleAdultUttLen="P0.00S" startTime="PT258.38S" endTime="PT261.14S" />

```

What your pipeline is doing is extracting the numbers inside conversationInfo="|EC|9|4|2|CIC|TIMR|FI|", in particular, the number 4 here indicates that this turn has increased the CTC to 4. What we do instead is count the TIMR and TIFR, which are the codes that appear in the same conversationInfo each time LENA increases the CTC.

So it *should* be the same, but if LENA has taught me anything is humility when facing such assumptions.

alecristia (approved):

ChildProject/pipelines/metricsFunctions.py
LoannPeurey: We did not check thoroughly how the lena etraction was done and we will indeed love to have your view on those metrics. Just note that here it is ` CVC` and not ` CTC` .

ChildProject/tables.py
LoannPeurey: You are right, it does not need to be overridden, the exception was originally taking multiple arguments and was in the end simplified. So removing the __init__ makes sense. (and killing two birds with one stone, this will remove the unsual ` error` argument).

LoannPeurey:
> * A few places where I simply didn't understand what was happening.
> * Your way to calculate lena_CTC is different from mine. I'll check the difference in the results - if any - on Monday.
> 
> Other than that, everything seems good.

The code could probably benefit from more inline comments
- [ ] add comments

ChildProject/pipelines/metricsFunctions.py
kalenkovich: I guess I can't even read :-) 

To get CTC using already extracted columns, we could count rows where `lena_block_type` is one of AICF, AICM, CIC, and XIC.

I am not sure about CVC though: 
- I am not sure counting TIFR and TIMR yields CVC,
- some .its segments contain multiple vocalizations and the `childUttCnt` attribute has to be consulted in this case.

Here is an example that illustrates both points:

```<Segment spkr="CHN" average_dB="-31.94" peak_dB="-22.55" conversationInfo="|BC|19|1|0|CM|NT|FI|" childUttCnt="2" childUttLen="P0.34S" startUtt1="PT1182.79S" endUtt1="PT1183.05S" startUtt2="PT1183.40S" endUtt2="PT1183.48S" childCryVfxLen="P0.00S" startTime="PT1182.79S" endTime="PT1183.62S" />```

**Comment by LoannPeurey** [comment]
Added lena_CTC as you suggested (count of AICF AICM CIC XIC lines)

We will merge the branch but only release a new version after @kalenkovich 's testing of the segments extraction and lena CTC/CVC.

**Comment by kalenkovich** [comment]
I got the merge notification but missed your comment, @LoannPeurey. I'll try to check the calculation today.

What do you mean by "segments extraction"?

**Comment by kalenkovich** [comment]
I wasn't able to run the metrics pipeline. Here is what I tried:

```sh
child-project metrics /Users/ek221/blab/GIN/bergelson childproject_lena-metrics.csv custom metrics_list.csv --by recording_filename --recordings recordings.csv --period 5mins
```

And here is what I got:

```sh
child-project: error: unrecognized arguments: --by recording_filename --recordings recordings.csv --period 5mins
```

If I remove those three options, it works. But not the way I intended it to 😄  My best guess is that I messed up the order of the options/arguments. @LoannPeurey, do you mind checking what I did wrong?

And [here](https://github.com/BergelsonLab/one_time_scripts/tree/ek_cp_lena_metrics_test/20220819_childproject_lena-ctc-cvc) are the recordings and metrics file I used.

LoannPeurey:
Sorry for the delay, I am currently on leave until next monday so only checking sporadically email/github.
I suggest moving the options before the `custom` keyword.
So that should give
```bash
child-project metrics /Users/ek221/blab/GIN/bergelson childproject_lena-metrics.csv --by recording_filename --recordings recordings.csv --period 5mins custom metrics_list.csv
```
This specificity of the order feels like unwanted complexity. I should explore options.
In the meantime, try checking the [doc page](https://childproject.readthedocs.io/en/latest/metrics.html) and putting all the arguments belonging to the base `metrics` pipeline before the keyword (`lena`,`custom`, etc) and any argument belonging to a specific keyword subcommand after.

LoannPeurey:
> I got the merge notification but missed your comment, @LoannPeurey. I'll try to check the calculation today.
> 
> What do you mean by "segments extraction"?

What I mean is extracting metrics from a list of audio fragments defined by recording_filename, segment_onset, segment_offset

To do that, you must use a csv (eg segments.csv) file with those 3 columns along with the CLI call:
```bash
child-project metrics /Users/ek221/blab/GIN/bergelson childproject_lena-metrics.csv --by segments --segments segments.csv custom metrics_list.csv
```

kalenkovich:
Got it, thank you! Enjoy your time off.

PR #396, opened 2022-08-12
Annotations/merge single annot

LoannPeurey:
Allow merging to be done with increments by having:
- `skip_existing` option skipping all the existing converted files
- `recording_filter` merging only the lines belonging to the recording_filename values  in this set
- `full_merge_set` Allowing to merge even if the set already exists

### DIFF
--- a/ChildProject/annotations.py
+++ b/ChildProject/annotations.py
@@ -90,6 +90,12 @@ class AnnotationManager:
             required=False,
             generated=True,
         ),
+        IndexColumn(
+            name="merged_from",
+            description="sets used to generate this annotation by merging (comma separated)",
+            required=False,
+            generated=True,
+        ),
     ]
 
     SEGMENTS_COLUMNS = [
@@ -703,7 +709,7 @@ class AnnotationManager:
         self.write()
 
     def merge_annotations(
-        self, left_columns, right_columns, columns, output_set, input
+        self, left_columns, right_columns, columns, output_set, input, skip_existing: bool = False
     ):
         """From 2 DataFrames listing the annotation indexes to merge together (those indexes should come from
         the intersection of the left_set and right_set indexes), the listing of the columns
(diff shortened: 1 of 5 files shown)


LoannPeurey:
I am wondering if we should add CLI options to use those new possibilities with the merge? It seems like a rare occurence and adding a lot of options can be confusing. However, a number of datasets will be adding data incrementaly

ChildProject/annotations.py
```diff
@@ -703,7 +709,7 @@ def rename_set(
         self.write()
 
     def merge_annotations(
-        self, left_columns, right_columns, columns, output_set, input
+        self, left_columns, right_columns, columns, output_set, input, skip_existing: bool = False
```
alecristia, line 712: so by default it will overwrite, correct?

tests/test_annotations.py
```diff
@@ -262,19 +263,45 @@ def test_merge(project):
...
+        full_set_merge = False,
+        recording_filter = {'sound.wav'}
     )
     am.read()
 
+    anns = am.annotations[am.annotations['set'] == 'alice_vtc']
+    assert anns.shape[0] == 1
+    assert anns.iloc[0]['recording_filename'] == 'sound.wav'
+    
+    time.sleep(2) #sleeping for 2 seconds to have different 'imported_at' values so that can make sure both merge did fine
```
alecristia, line 287: nice trick!

alecristia (commented):
I'm having a hard time following this file, not sure if it's the time or what... so I hope you can get another set of eyes on this. Otherwise, may I suggest to delay the integration until WH is back, so he can at least test as soon as it's integrated, & we roll back if something went wrong?

ChildProject/annotations.py
LoannPeurey: Yes, I think by default it is better than a 'merge' re-generates the converted annotations, this will update the merged set if there were changes to the original sets. So not updating the existing converted files should be a deliberate choice by the user.

LoannPeurey:
> I'm having a hard time following this file, not sure if it's the time or what... so I hope you can get another set of eyes on this. Otherwise, may I suggest to delay the integration until WH is back, so he can at least test as soon as it's integrated, & we roll back if something went wrong?

We could consider merging into master but not releasing until William tests it. How does that sound?

tests/test_annotations.py
LoannPeurey: The automated tests are fasr =/ . I need to make sure the importation was indeed only done on the second file and skipped the existing one. But this does not look like an elegant solution. I will try to think about a better one

PR #411, opened 2023-01-03
Check calculated ages during corpus validation

lucasgautheron:
Recently there was a date of birth issue with Fausey-trio, such that some of the ages calculated from the dates of the recordings and the dates of birth were negative. Corpus validation should compute ages and make sure they are all positive at least.

This PR does just that.


lucasgautheron:
Weird indeed! I re-ran the failed tests and it worked this time!

PR #413, opened 2023-02-08
Annotations/enforce unicity

LoannPeurey:
The purpose of the PR is to enforce that:
- we cannot have multiple lines in annotations.csv with the same set and annotation_filename
- A set cannot have overlapping annotation files. Meaning inside a specific set, it is not possible to annotate multiple times the same stretch of audio (bc then this stretch is counted multiple times in metrics and other analysis)

To achieve this, all the functions that interact with annotations.csv must make sure before writing to it that those cases don't exist,
this is especially relevant for importations that now will fail for every row that does not satisfy those conditions.

Previously, the errors encountered during importation would find themselves recorded into annotations.csv. I felt like this could be troublesome as the file is intended as a record of indexed annotation and should not serve as a logfile. My take was to output all the errors to a separate csv in extra but not too sure about this solution

### DIFF
--- a/ChildProject/annotations.py
+++ b/ChildProject/annotations.py
@@ -463,7 +463,8 @@ class AnnotationManager:
             if overwrite_existing:
                 print(f"Warning: annotation file {output_filename} will be overwritten")
             else:
-                annotation["error"] = f"annotation file {output_filename} already exists, to reimport it, use the overwrite_existing flag"
+                error_filename = output_filename.replace('\\','/')
+                annotation["error"] = f"annotation file {error_filename} already exists, to reimport it, use the overwrite_existing flag"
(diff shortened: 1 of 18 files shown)


LoannPeurey:
I am also unsatisfied about the
https://github.com/LAAC-LSCP/ChildProject/blob/6f6474f0dafb980a88d4e16b62b36f57cf23e7fe/ChildProject/utils.py#L180 function to handle correctly when we want to select by a number of other columns, I could not think immediately of a clean way to do it without `eval()` . So if you have suggestions..

William-N-Havard (approved):

shuvayanti (approved):
Everything looks good to me. But I see a merge conflict which probably need to be resolved.

PR #415, opened 2023-02-28
Projects/fix discard

LoannPeurey:
fix the usage of discard in datasets
change the edition of children or recordings record by having the possibility of keeping lines and columns discarded

### DIFF
--- a/ChildProject/cmdline.py
+++ b/ChildProject/cmdline.py
@@ -517,9 +517,6 @@ def compute_durations(args):
     
     project.recordings = recordings.copy()
     project.write_recordings()
-#    recordings.to_csv(
-#        os.path.join(project.path, "metadata/recordings.csv"), index=False
-#    )
     
 @subcommand(
     [
--- a/ChildProject/cmdline.py
+++ b/ChildProject/cmdline.py
@@ -513,10 +513,13 @@ def compute_durations(args):
         right_on="recording_filename",
     )
     recordings["duration"].fillna(0, inplace=True)
-    recordings["duration"] = recordings["duration"].astype(int)
-    recordings.to_csv(
-        os.path.join(project.path, "metadata/recordings.csv"), index=False
-    )
+    recordings["duration"] = recordings["duration"].astype("Int64")
+    
+    project.recordings = recordings.copy()
+    project.write_recordings()
+#    recordings.to_csv(
+#        os.path.join(project.path, "metadata/recordings.csv"), index=False
+#    )
     
 @subcommand(
     [
--- a/ChildProject/projects.py
+++ b/ChildProject/projects.py
@@ -16,6 +16,17 @@ from .tables import (
 )
 from .utils import get_audio_duration, path_is_parent
 
+RAW_RECORDINGS = os.path.normpath("recordings/raw")
+CONVERTED_RECORDINGS = os.path.normpath("recordings/converted")
+STANDARD_SAMPLE_RATE = 16000
+STANDARD_PROFILE = 'standard' # profile that is expected to contain the standardized audios (16kHz). The existence and sampling rates of this profile are checked when <validating this profile> or <validating without profile and the raw recordings are not 16kHz>.
+
+METADATA_FOLDER = 'metadata'
+CHILDREN_CSV = 'children.csv'
+RECORDINGS_CSV = 'recordings.csv'
+
+PROJECT_FOLDERS = ["recordings", "annotations", "metadata", "doc", "scripts"]
+
 
 class ChildProject:
     """ChildProject instance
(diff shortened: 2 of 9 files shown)


ChildProject/cmdline.py
```diff
@@ -513,10 +513,13 @@ def compute_durations(args):
...
-    recordings.to_csv(
-        os.path.join(project.path, "metadata/recordings.csv"), index=False
-    )
+    recordings["duration"] = recordings["duration"].astype("Int64")
+    
+    project.recordings = recordings.copy()
+    project.write_recordings()
+#    recordings.to_csv(
+#        os.path.join(project.path, "metadata/recordings.csv"), index=False
+#    )
```
alecristia, line 522: is it typically ok to leave old code commented out? or is it better to remove it?

ChildProject/projects.py
```diff
@@ -16,6 +16,17 @@
...
+RAW_RECORDINGS = os.path.normpath("recordings/raw")
+CONVERTED_RECORDINGS = os.path.normpath("recordings/converted")
+STANDARD_SAMPLE_RATE = 16000
+STANDARD_PROFILE = 'standard' # profile that is expected to contain the standardized audios (16kHz). The existence and sampling rates of this profile are checked when <validating this profile> or <validating without profile and the raw recordings are not 16kHz>.
+
+METADATA_FOLDER = 'metadata'
+CHILDREN_CSV = 'children.csv'
+RECORDINGS_CSV = 'recordings.csv'
+
+PROJECT_FOLDERS = ["recordings", "annotations", "metadata", "doc", "scripts"]
```
alecristia, line 28: I'm noticing all of these variables are in all caps. Is that what is done elsewhere in the package? does it have a specific meaning? is it better this way or a different way?

alecristia: ps great idea to use normpath - is that what we do elsewhere?

ChildProject/projects.py
```diff
@@ -27,7 +38,7 @@ class ChildProject:
     :type path: str
     :param enforce_dtypes: enforce dtypes on children/recordings dataframes, defaults to False
     :type enforce_dtypes: bool, optional
-    :param ignore_discarded: ignore entries such that discard=1, defaults to False
+    :param ignore_discarded: ignore entries such that discard=1, defaults to True
```
alecristia, line 41: wow! big change of default!

ChildProject/projects.py
```diff
@@ -290,7 +296,7 @@ def accumulate_metadata(
         merge_column: str,
         verbose=False,
     ) -> pd.DataFrame:
-        md_path = os.path.join(self.path, "metadata", table)
+        md_path = os.path.join(self.path, METADATA_FOLDER, table)
```
alecristia, line 299: ok this resolves to the same thing, since md_folder defined above - any reason to make it into a variable? should we do the same elsewhere?

tests/test_validation.py
```diff
@@ -8,7 +8,7 @@ def test_valid_project():
     errors, warnings = project.validate()
 
     assert len(errors) == 0, "valid input validation failed (expected to pass)"
-    assert len(warnings) == 3, "expected 1 warning, got {}".format(len(warnings))
+    assert len(warnings) == 4, "expected 3 warning, got {}".format(len(warnings))
```
alecristia, line 11: minor: expected 3 warningS

ChildProject/cmdline.py
LoannPeurey: I think sometimes it is useful if the old code has a different approach and knowing the original helps.
But here clearly not, this is quite dirty way of writing the file and it is not good to keep it there imo.

ChildProject/projects.py
LoannPeurey: All caps is reserved for fixed variables, i a way to indicate that you can change easily entire aspects of the package by changing those variables without breaking anything. My change here:
- for those existing, take them out of the class and put them in all caps as they should not be built each time we create a project object, it is better practice to have them as constants at the level of the module (file)
- for those not existing, replace hard coded strings in the code by them to make changes easier. This way if one day we change 'children.csv' with 'particpants.csv' , it should just be a matter of changing that constant and everything should work (probably not yet as there is a lot of harmonization to be done on the whole package)

LoannPeurey: I started to use normpath in an effort to have automated windows testing (which we do have and maintain now). There are some areas of the package where normpath is not really integrated correctly to support all of that but they work because libraries are smart enough to use linux paths in windows. But it would definitely be better to make that the norm

ChildProject/projects.py
LoannPeurey: Yes and no, the usage of the `discard` flag is really scarce into our current datasets. Especially as the existing code was bugged so basically it was behaving opposite of what it should have done. The fix to this and that change of default behavior will hopefully open the door to using more consistently the discard flag to ignore edge case audios (like tests, too small etc) rather than removing them from the dataset and losing all trace of their existence.

ChildProject/projects.py
LoannPeurey: We should definitely do this as much as we can. This brings more flexibility in changing the code like explained above

tests/test_validation.py
LoannPeurey: :scream_cat:

ChildProject/projects.py
alecristia: sounds great - one more thing to add to the "todo" list for may-july!

alecristia (approved):
i got it now! a rest gives clarity

ChildProject/cmdline.py
LoannPeurey: removed

PR #429, opened 2023-04-27
Zooniverse uploads and robustness

LoannPeurey:
Make zooniverse uploads more robust and record additional metadata


LoannPeurey:
ChildProject proojects were changed to enforce unicity in the experiment column and have an 'experiment' value.
an automated test should be added

LoannPeurey:
Some handling of the 'maximum number of subjects were uploaded for that account' must be added to make sure we don't keep trying to upload for nothing even with the ignore errors option

shuvayanti (approved):

Issue #430, opened 2023-04-28
get_within_time_range interface


lucasgautheron:
(I thought I posted this issue already, but apparently not -- can't find it!)

I have two questions regarding the new parametrization of AnnotationManager.get_within_time_range(), from strings to a TimeInterval object.

1. The documentation does not reflect the change (the example here is outdated: https://childproject.readthedocs.io/en/latest/api-annotations.html#annotations-within-a-specific-time-range)
2. Is it possible to have something closer to the previous interface? I find the new parametrization confusing for two reasons indeed:
  - First, it requires to import two additional objects: TimeInterval and datetime
  - Second, it requires the specification of a full datetime object, even though we only care about time of the day. This makes it unnecessarily complicated and also introduces confusion about the behavior of the function: it makes it unclear whether the functions filters annotations between two dates _and_ times, or instead annotations between two hours of the day (regardless of the date)

LoannPeurey:
I think you make a good point, the function was changed when reworking the metrics system, the values in the dataframes are datetime objects and forcing the function to use str values meant that Everything was converted to str, and converted back to datetime, which is not ideal.

So I think I underestimated the ease of use and clarity of the function when used externally.

I think it would make sense to keep both possibilities of using the function by using `start_time` and `end_time` with `interval` at None

It would mean that using the function with str value would be :
`am.get_within_time_range(am.annotations, start_time='09:00', end_time='12:00')`

What do you think? I opened a branch and PR #431 to work on this.

lucasgautheron:
I think that's a very good idea!

PR #431, opened 2023-04-28
Annotations/time ranges with str

LoannPeurey:
#430


lucasgautheron (commented):
Is the `errors`  argument (default value: raise) used somewhere? might be something I forgot to actually implement back then. Otherwise that seems good to me :)

PR #478, opened 2024-06-26
Conda Mac Python version upgrade

tsepuri:
Supporting ARM Macs by default without Rosetta


LoannPeurey:
thank you, yes that should have been done a long time ago, like for the linux .yml

PR #501, opened 2025-06-30
bug fixes

LoannPeurey:
fixing 2 bugs

### DIFF
--- a/ChildProject/projects.py
+++ b/ChildProject/projects.py
@@ -1,3 +1,4 @@
+import sys
 import datetime
 from functools import partial
 import logging
@@ -6,7 +7,11 @@ import os
 import pandas as pd
 import re
 import shutil
-from typing import Union, Self
+from typing import Union, List, Tuple, Optional
+if sys.version_info[0] == 3 and sys.version_info[1] >= 11:
+    from typing import Self
+else:
+    from typing_extensions import Self
 from pathlib import Path
 
 from .tables import (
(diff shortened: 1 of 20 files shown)


ChildProject/projects.py
```diff
@@ -504,6 +508,8 @@ def remove_project_file(self, file, file_type: str):
 
         destination.unlink()
 
+        return self
```
LawrenceBorst, line 511: I guess it's just a matter of taste whether we should return the instance? The nice part is that we don't have to think about whether a function changes state/has side effects or not, since we can always just assign. I had to double check the cost of returning, but it's negligible, even with assignment because a class instance would be assigned to itself

I do think if we want to do this we should also type-hint the return value in the function signature. Tools like IntelliSense need this for typing info if a variable ever gets reassigned to the return value a function like this one

LawrenceBorst:
Should we consider doing merges with rebase instead of merging pure and simple 🤔 in general? Many projects prefer it that way, it's often easy to understand, and kind of forces you to resolve problems in interactive rebase

ChildProject/projects.py
LoannPeurey: updated type ints in [a671b36](https://github.com/LAAC-LSCP/ChildProject/pull/501/commits/a671b369bba5f2975410692b80c32242b71a6610)

tests/test_cli.py
```diff
@@ -50,6 +50,14 @@ def dataset_setup(request):
 
     yield 1
 
+# def segments():
```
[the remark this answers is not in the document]
LoannPeurey: removed :+1:

PR #503, opened 2025-07-16
Fix decorator not returning function

LawrenceBorst:
Small fix. Need it also for exporting command line functions properly (otherwise show up as None in the module object)


LawrenceBorst:
I see this is already covered by Loann's various bug fixes in #501 👀

